PR #10, opened 2016-07-11
Adds Redis-backed storage mechanism for UserPermissions

ttomsu:
Some Jedis classes copy/pasted from other components. 

@cfieber @ajordens @jtk54 @duftler PTAL


fiat-core/src/main/java/com/netflix/spinnaker/fiat/model/resources/Resource.java
```diff
@@ -0,0 +1,36 @@
...
+
+import lombok.NonNull;
+import org.apache.commons.lang3.StringUtils;
+
+public enum Resource {
+  ACCOUNT,
+  APPLICATION;
+
+  // TODO(ttomsu): This is Redis-specific, so it probably shouldn't go here.
+  public static Resource parse(@NonNull String key) {
```
jtk54: Maybe in a `Utils` class?

fiat-roles/src/main/groovy/com/netflix/spinnaker/fiat/permissions/RedisPermissionsRepository.java
```diff
@@ -16,7 +16,228 @@
...
+    }
+    return allById;
+  }
+
+  private Table<String, Resource, Response<Map<String, String>>> getAllFromRedis() {
+    try (Jedis jedis = jedisSource.getJedis()) {
+      Set<String> allUserIds = jedis.smembers(allUsersKey());
+
+      Table<String, Resource, Response<Map<String, String>>> responseTable =
+          HashBasedTable.create(allUserIds.size(), Resource.values().length);
```
jtk54: Why `HashBasedTable` here? To make a flat table instead of a nested `HashMap`?

jtk54:
LGTM, deferring to the other PTALs to merge : D

fiat-roles/src/main/groovy/com/netflix/spinnaker/fiat/permissions/RedisPermissionsRepository.java
jtk54: Guava might be a chunky dependency if we don't use it anywhere else.

fiat-roles/src/test/groovy/com/netflix/spinnaker/fiat/redis/EmbeddedRedis.java
```diff
@@ -0,0 +1,77 @@
...
+import java.net.URISyntaxException;
+
+public class EmbeddedRedis {
+
+  private final URI connection;
+  private final RedisServer redisServer;
+
+  private JedisPoolSource poolSource;
+
+  private EmbeddedRedis(int port) throws IOException, URISyntaxException {
```
cfieber: any particular reason for this as opposed to a dependency on [kork-jedis-test](https://github.com/spinnaker/kork/tree/master/kork-jedis-test) ?

fiat-roles/src/main/groovy/com/netflix/spinnaker/fiat/config/RedisConfig.groovy
```diff
@@ -0,0 +1,117 @@
...
+import org.springframework.context.annotation.Configuration
+import redis.clients.jedis.Jedis
+import redis.clients.jedis.JedisPool
+import redis.clients.jedis.Protocol
+
+@Configuration
+@ConditionalOnProperty("redis.connection")
+class RedisConfig {
+
+  @Value('${redis.scheduler: default}')
```
cfieber: think these were the config keys for the cloud driver caching agent scheduler?

fiat-roles/src/main/groovy/com/netflix/spinnaker/fiat/config/RedisConfig.groovy
```diff
@@ -0,0 +1,117 @@
...
+  String schedulerType;
+
+  @Value('${redis.parallelism: -1}')
+  Integer parallelism;
```
cfieber: same as above

fiat-roles/src/main/groovy/com/netflix/spinnaker/fiat/config/RedisConfig.groovy
```diff
@@ -0,0 +1,117 @@
...
+
+  @Bean
+  JedisSource jedisSource(JedisPool jedisPool) {
+    return new JedisPoolSource(jedisPool);
+  }
+
+  @Bean
+  @ConfigurationProperties('redis')
+  GenericObjectPoolConfig redisPoolConfig() {
+    new GenericObjectPoolConfig(maxTotal: 100, maxIdle: 100, minIdle: 25)
```
cfieber: the sizing on this pool is probably larger than needed for this service FWIW

fiat-roles/src/main/groovy/com/netflix/spinnaker/fiat/config/RedisConfig.groovy
```diff
@@ -0,0 +1,117 @@
...
+
+    int database = Integer.parseInt((redisConnection.path ?: "/${Protocol.DEFAULT_DATABASE}").split('/', 2)[1])
+
+    String password = redisConnection.userInfo ? redisConnection.userInfo.split(':', 2)[1] : null
+
+    new JedisPool(redisPoolConfig ?: new GenericObjectPoolConfig(), host, port, timeout, password, database, null)
+  }
+
+  @Bean
+  JedisPool jedisPoolPrevious(
```
cfieber: this was from clouddriver to support reading from an old redis to enable a redeploy that changed the key format - probably not needed in here

fiat-roles/fiat-roles.gradle
```diff
@@ -15,14 +15,19 @@
...
-  compile project(':fiat-core')
+  compile project(":fiat-core")
 
   spinnaker.group("retrofitDefault")
 
-  compile spinnaker.dependency('bootWeb')
-  compile spinnaker.dependency('groovy')
+  compile spinnaker.dependency("bootActuator")
+  compile spinnaker.dependency("bootWeb")
+  compile spinnaker.dependency("groovy")
```
cfieber: feels like this came in just to bring in `RedisConfig.groovy` unchanged.

I don't think there is actually that much in there once the unneeded bits get cut out that wouldn't be too hard to port over to java (aside from the one `//thx groovy` where it's reading a private field)

fiat-roles/src/main/groovy/com/netflix/spinnaker/fiat/permissions/RedisPermissionsRepository.java
```diff
@@ -16,7 +16,228 @@
...
+          case ACCOUNT:
+            permission.getAccounts().stream().forEach(putInValuesMap);
+            break;
+          case APPLICATION:
+            permission.getApplications().stream().forEach(putInValuesMap);
+            break;
+        }
+
+        if (!resourceValues.isEmpty()) {
+          jedis.hmset(userKey(permission.getId(), r), resourceValues);
```
cfieber: could theoretically do `.pipelined()` before the `for (Resource r`, and do these calls on the pipeline, then commit it at after the end of the for loop

fiat-roles/src/main/groovy/com/netflix/spinnaker/fiat/permissions/RedisPermissionsRepository.java
```diff
@@ -16,7 +16,228 @@
...
+          case ACCOUNT:
+            userPermission.getAccounts().addAll(extractAccounts(resourceMap));
+            break;
+          case APPLICATION:
+            userPermission.getApplications().addAll(extractApplications(resourceMap));
+            break;
+        }
+      }
+    } catch (Exception e) {
+      log.error("Storage exception reading " + id + " entry.", e);
```
cfieber: should this throw something on read error rather than returning null? if it should return null should it return `Optional<UserPermission>` instead?

cfieber:
The main question I have is around the behaviour of the repository wrt persistence exceptions - I think logging and returning nulls and empties probably isn't the right approach. Throwing so the caller can implement retries (or pushing retries down into the repository) seems better.

ajordens:
Part of me questions whether Redis provides the right level of durability if `fiat` is the system of record for permissions ... thoughts? That's the intent right, or am I mistaken.

fiat-roles/src/main/groovy/com/netflix/spinnaker/fiat/permissions/RedisPermissionsRepository.java
ajordens: I'd probably go with Optional<UserPermission> and forgo the `Optional.ofNullable()` in the callers.

fiat-roles/src/test/groovy/com/netflix/spinnaker/fiat/permissions/RedisPermissionsRepositorySpec.groovy
```diff
@@ -0,0 +1,161 @@
...
+        .setJedisSource(embeddedRedis.jedisPoolSource)
+  }
+
+  def cleanup() {
+    jedis.flushDB()
+  }
+
+  def "should put the specified permission in redis"() {
+    setup:
+      Account account1 = new Account().setName("account1")
```
ajordens: (minor) but no need to indent

fiat-core/src/main/java/com/netflix/spinnaker/fiat/model/resources/Resource.java
ttomsu: I'm willing to leave it here for now, just left the TODO as a reminder to maybe find a better place for it in the future, if it ends up being necessary.

fiat-roles/fiat-roles.gradle
ttomsu: The thing that changed was the quote style - single to double. Groovy was pulled in when the role syncing stuff was copied over from Gate a few PRs ago. I plan on rewriting those bits to Java 8, but that's more of a cleanup task and I wanted to get everything functioning first.

fiat-roles/src/main/groovy/com/netflix/spinnaker/fiat/permissions/RedisPermissionsRepository.java
ttomsu: I like it.

fiat-roles/src/main/groovy/com/netflix/spinnaker/fiat/permissions/RedisPermissionsRepository.java
ttomsu: If we find the binary is too large we can work to slim it down, but lets not prematurely optimize.

fiat-roles/src/test/groovy/com/netflix/spinnaker/fiat/permissions/RedisPermissionsRepositorySpec.groovy
ttomsu: True, but I find it easier to read when indented from the the labels.

ajordens: Lets follow the conventions from other the other projects and not-indent.

fiat-roles/src/test/groovy/com/netflix/spinnaker/fiat/redis/EmbeddedRedis.java
ttomsu: Because I didn't know about it. It appears we can remove the copy of EmbeddedRedis from Gate by doing the same (@jtk54)

ttomsu:
Ready again for review.

ttomsu:
@ajordens All permissions should be recreateable if Fiat were to go down. Users may have to log in again if accessing protected resources (if, in your case, the groups come along with the SAML assertion). 

@cfieber Retries and more advanced exception handling can definitely be added. In fact, there is a TODO in the `PermissionRepository` interface that mentions that this access pattern may need to be optimized once we get into into integrating the other components.

ajordens:
@ttomsu Cool ... so the deal is that the subjects (ie. applications/accounts) requiring authentication will be periodically seeded into `fiat` from 1..* other systems? 

Hypothetically `clouddriver` supplying auth requirements at the account level and `front50` supplying auth requirements at the application level?

We've started to have a few conversations internally with platform & application sec ... they're curious because they see a need for a similar can `a` do `b` on `c` kind of system outside of Spinnaker.

From my perspective, I just want to make sure we're not storing any durable data in the `fiat` redis that can't be re-seeded ... and really the `redis` is an optimization towards consistency with multiple `fiats` running in an environment.

ttomsu:
@ajordens Accounts are statically defined in YAML config files - just like they are in Clouddriver today. The declaration of `requiredGroupMembership` will need to move over to the `fiat-local.yml` config file. 

Applications will be more dynamic (see #11), fed in from an external persistence mechanism. I plan on adding GCS & S3 bucket implementations, to be consistent with the pipeline config storage mechanism.

You're correct that durable data is not being stored in Redis. Account config is static in config files, and dynamic application config is in external persistence. Shutting down Fiat, flushing Redis, and restarting Fiat will cause all logged in users to lose access to protected resources, so they would need to logout + login again.

fiat-roles/src/main/groovy/com/netflix/spinnaker/fiat/permissions/RedisPermissionsRepository.java
ttomsu: A change in the next PR introduces a delete call to fix a bug, and would obviate the `pipeline` addition.

ajordens:
@ttomsu It would be nice to not require duplication of account details between `clouddriver` and `fiat` IMO ... not sure what it buys us otherwise?

What is the motivation around not storing the auth details in `front50`? I suspect when it comes to doing GCS/S3, you would use the same buckets anyways. 

It might be simpler to have a `Provider`-like interface with a couple implementations ... file-based to ease testing and initial development (like a mock), and another to support production where the information is sourced from other systems of record. 

Happy to hear arguments to the contrary. Cheers!

ttomsu:
Spoke with @ajordens and @cfieber offline - we're going to push through with this and the other PRs (#11, #12) now, and replace the configuration implementations with calls to Front50 and Clouddriver for application and account configurations, respectively.

PR #11, opened 2016-07-12
Adds ApplicationProvider interface and file watching implementation.

ttomsu:
Based off of #10 , so please only review the **last** commit.

The intent is to get these app configs moved into an S3/GCS bucket, but starting with a file-based approach helps prove out the concepts.

A user's `fiat-local.yml` file would contain something like:
```
applications:
  fileBased:
    directory: /home/ttomsu/.spinnaker/appConfigs
```
where admins of the system would drop in files like `my-app.config` that look similar to how account configs are structured:
```
name: my-app
requiredGroupMembership:
- my-app-users
```

Eventually we'll get to the point where these files will be generated when an app is created in the UI and uploaded into the configured storage bucket, where it can be versioned and watched for changes by all Fiat instances. 

@cfieber @ajordens @duftler @jtk54 PTAL

### DIFF
--- a/fiat-core/src/main/java/com/netflix/spinnaker/fiat/providers/ApplicationProvider.java
+++ b/fiat-core/src/main/java/com/netflix/spinnaker/fiat/providers/ApplicationProvider.java
@@ -14,10 +14,14 @@
  * limitations under the License.
  */
 
-package com.netflix.spinnaker.fiat.model;
+package com.netflix.spinnaker.fiat.providers;
 
-public enum Authorization {
-  CREATE,
-  READ,
-  WRITE
+import com.netflix.spinnaker.fiat.model.resources.Application;
+
+import java.util.Collection;
+import java.util.Set;
+
+public interface ApplicationProvider {
+
+  Set<Application> getApplications(Collection<String> groups);
 }
--- a/fiat-core/src/main/java/com/netflix/spinnaker/fiat/providers/FileBasedApplicationProvider.java
+++ b/fiat-core/src/main/java/com/netflix/spinnaker/fiat/providers/FileBasedApplicationProvider.java
@@ -1,0 +1,172 @@
+/*
+ * Copyright 2016 Google, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.fiat.providers;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.dataformat.yaml.YAMLFactory;
+import com.google.common.annotations.VisibleForTesting;
+import com.netflix.spinnaker.fiat.model.resources.Application;
+import lombok.NonNull;
+import lombok.RequiredArgsConstructor;
+import lombok.Setter;
+import lombok.extern.slf4j.Slf4j;
+import lombok.val;
+import org.apache.commons.lang3.StringUtils;
+
+import javax.annotation.PreDestroy;
+import java.io.IOException;
+import java.nio.file.*;
+import java.util.*;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.stream.Collectors;
+
+@Slf4j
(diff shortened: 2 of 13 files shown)


fiat-core/src/main/java/com/netflix/spinnaker/fiat/providers/ApplicationProvider.java
```diff
@@ -0,0 +1,28 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.fiat.providers;
+
+import com.netflix.spinnaker.fiat.model.resources.Application;
+import lombok.NonNull;
```
jtk54: This import isn't used.

fiat-core/src/main/java/com/netflix/spinnaker/fiat/providers/FileBasedApplicationProvider.java
```diff
@@ -0,0 +1,171 @@
...
+            WatchKey watchKey = watchService.take();
+
+            for (WatchEvent event : watchKey.pollEvents()) {
+              String filepath = event.context().toString();
+              if (!isConfigFile(filepath)) {
+                continue;
+              }
+
+              switch (event.kind().toString()) {
+                case "ENTRY_CREATE":
```
jtk54: I think we'd like to create new applications and have them sync into Fiat without restarting or explicitly reloading from the config path.

fiat-core/src/main/java/com/netflix/spinnaker/fiat/providers/FileBasedApplicationProvider.java
```diff
@@ -0,0 +1,171 @@
...
+                case "ENTRY_MODIFY":
+                  val fullConfigPath = watchedDirPath.resolve((Path) event.context());
```
jtk54: A `log.info` here would probably be helpful, same with `ENTRY_CREATE` if you implement that one.

fiat-roles/src/main/groovy/com/netflix/spinnaker/fiat/permissions/RedisPermissionsRepository.java
```diff
@@ -98,8 +98,10 @@ public PermissionsRepository put(@NonNull UserPermission permission) {
             break;
         }
 
+        String userResourceKey = userKey(permission.getId(), r);
```
jtk54: `permission.getId()` is called twice, could use a local variable.

fiat-core/src/main/java/com/netflix/spinnaker/fiat/providers/FileBasedApplicationProvider.java
ttomsu: That's what this does. It's watching a specified directory for new or modified files.

fiat-core/src/main/java/com/netflix/spinnaker/fiat/providers/FileBasedApplicationProvider.java
ttomsu:
```
case "ENTRY_CREATE":
case "ENTRY_MODIFY":
``` 

is a switch fall through - meaning execute whatever code is below up until the next `break` for either ENTRY_CREATE or ENTRY_MODIFY. Adding a log statement...

fiat-roles/src/main/groovy/com/netflix/spinnaker/fiat/permissions/RedisPermissionsRepository.java
ttomsu: Done.

fiat-core/src/main/java/com/netflix/spinnaker/fiat/providers/FileBasedApplicationProvider.java
jtk54: Got it, so `CREATE` and `MODIFY` are handled by the same case? I think a private function explicitly handling both cases with a `break` would be easier to read.

ttomsu:
As mentioned in #10, I'll push ahead with this PR, but will come back to add the primary functionality of calling Front50 to retrieve application config data, rather than rely on a separate watcher.

fiat-core/src/main/java/com/netflix/spinnaker/fiat/providers/FileBasedApplicationProvider.java
ttomsu: I'm going to have to disagree - switch fallthroughs are a well known "feature" of the language, and given that the actual amount of work is contained within 4 lines of code, I think adding two method calls would detract from the clarity.

That being said, this code will most likely be removed shortly, as the preferred way to get application data is now through Front50.

PR #12, opened 2016-07-13
Adds service account concept and static config setup, impl.

ttomsu:
This is based off of #11, please only review the **last** commit.

Introduces the Fiat service account - a mechanism to allow automated/triggered pipelines to access protected resources. 

Because it is itself a group, only users who belong to the service account group will be able to see that it exists and to use it. This is reflected in the returned UserPermission object:

```
ttomsu@blinky:~/.spinnaker/appConfigs$ curl http://localhost:7003/authorize/ttomsu@spinnaker-test.net | pjs
{
    "cloudResources": {
        "accounts": [
            {
                "authorizations": [
                    "READ",
                    "WRITE"
                ],
                "name": "my-protected-account"
            },
            {
                "authorizations": [
                    "READ",
                    "WRITE"
                ],
                "name": "my-aws-account"
            }
        ],
        "applications": [
            {
                "authorizations": [
                    "CREATE",
                    "READ",
                    "WRITE"
                ],
                "name": "abc"
            }
        ]
    },
    "name": "ttomsu@spinnaker-test.net",
    "serviceAccounts": [
        {
            "name": "sekret-service-acct"
        }
    ]
}
```

This means that service account can be thought of/queried as a normal, real user:
```
ttomsu@blinky:~/.spinnaker/appConfigs$ curl http://localhost:7003/authorize/sekret-service-acct@spinnaker-test.net | pjs

{
    "cloudResources": {
        "accounts": [
            {
                "authorizations": [
                    "READ",
                    "WRITE"
                ],
                "name": "my-protected-account"
            },
            {
                "authorizations": [
                    "READ",
                    "WRITE"
                ],
                "name": "my-aws-account"
            }
        ],
        "applications": [
            {
                "authorizations": [
                    "CREATE",
                    "READ",
                    "WRITE"
                ],
                "name": "abc"
            }
        ]
    },
    "name": "sekret-service-acct@spinnaker-test.net",
    "serviceAccounts": []
}
```

When a user (with access to the service account) adds a trigger to a pipeline, and configures that trigger to run the pipeline as a service account, that pipeline will execute as if a real user logged in and clicked "Run" - including all authorization checks as if it were the real user.

This impl uses the Spring config to manage service accounts, but it's likely that real users will want to add/remove them through the application itself. We will probably end up with a similar scenario with how application configs are managed. 

@cfieber @ajordens @jtk54 @duftler PTAL.

### DIFF
--- a/fiat-core/src/main/java/com/netflix/spinnaker/fiat/model/UserPermission.java
+++ b/fiat-core/src/main/java/com/netflix/spinnaker/fiat/model/UserPermission.java
@@ -32,6 +32,7 @@ public class UserPermission {
   private String id;
   private Set<Account> accounts = new HashSet<>();
   private Set<Application> applications = new HashSet<>();
+  private Set<ServiceAccount> serviceAccounts = new HashSet<>();
 
   @JsonIgnore
   public boolean isEmpty() {
@@ -46,7 +47,11 @@ public class UserPermission {
   @Data
   public class View {
     String name = UserPermission.this.id;
-    Map<String, Object> resources = ImmutableMap.of(
+    Set<ServiceAccount.View> serviceAccounts = UserPermission.this.serviceAccounts
+        .stream()
+        .map(ServiceAccount::getView)
+        .collect(Collectors.toSet());
+    Map<String, Object> cloudResources = ImmutableMap.of(
         "accounts",
         UserPermission.this.accounts
             .stream()
--- a/fiat-core/src/main/java/com/netflix/spinnaker/fiat/model/ServiceAccount.java
+++ b/fiat-core/src/main/java/com/netflix/spinnaker/fiat/model/ServiceAccount.java
@@ -14,24 +14,23 @@
  * limitations under the License.
  */
 
-package com.netflix.spinnaker.fiat.model.resources;
+package com.netflix.spinnaker.fiat.model;
 
-import lombok.NonNull;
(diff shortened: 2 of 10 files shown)


fiat-core/src/main/java/com/netflix/spinnaker/fiat/model/UserPermission.java
```diff
@@ -46,7 +47,11 @@ public View getView() {
   @Data
   public class View {
     String name = UserPermission.this.id;
-    Map<String, Object> resources = ImmutableMap.of(
+    Set<ServiceAccount.View> serviceAccounts = UserPermission.this.serviceAccounts
+        .stream()
+        .map(ServiceAccount::getView)
+        .collect(Collectors.toSet());
+    Map<String, Object> cloudResources = ImmutableMap.of(
```
jtk54: Maybe type this stronger than `Object`? I think we can have a general `ProtectedResource` superclass that things like `Application`, etc extend so we at least a bit of static typing.

jtk54:
LGTM after comment.

fiat-core/src/main/java/com/netflix/spinnaker/fiat/model/UserPermission.java
ttomsu: `View` classes are used solely for serialization, so I think the typing is far less of a concern.

PR #15, opened 2016-07-19
Converts static config account provider into dynamic calls to Clouddr…

ttomsu:
…iver.

@duftler @cfieber @jtk54 PTAL

@ajordens FYI


duftler:
LGTM.

fiat-web/src/main/java/com/netflix/spinnaker/fiat/controllers/AuthorizeController.java
```diff
@@ -35,6 +40,26 @@
   @Autowired
   private PermissionsRepository permissionsRepository;
 
+  @Value("${auth.getAll.enabled:false}")
+  @Setter
+  private boolean getAllEnabled;
+
+  @ApiModelProperty(value = "Used mostly for testing. Not really any real value to the rest of " +
```
duftler: I think this is the wrong type of annotation here, no?

ttomsu: Fixed.

jtk54:
LGTMA

PR #16, opened 2016-07-19
Converts static application config to use Front50. Makes ServiceAccou…

ttomsu:
…ntProvider consistent with Account and ApplicationProviders.

@duftler @jtk54 @cfieber PTAL

@ajordens FYI


jtk54:
LGTM

PR #17, opened 2016-07-20
Converts static service account config to use Front50.

ttomsu:
Relies on https://github.com/spinnaker/front50/pull/105, but can be merged independently since this is component is not in service yet. 

@duftler @jtk54 @cfieber PTAL

@ajordens FYI


fiat-core/src/main/java/com/netflix/spinnaker/fiat/providers/DefaultServiceAccountProvider.java
```diff
@@ -16,40 +16,71 @@
...
+   * after the "@" symbol for the purposes of service account/group matching.
    */
   @Override
   public Set<ServiceAccount> getAccounts(@NonNull Collection<String> groups) {
-    if (serviceAccountsByName == null) {
-      return Collections.emptySet();
-    }
+    // There is a potential here for a naming collision where service account
+    // "my-svc-account@abc.com" and "my-svc-account@xyz.com" each allow one another's users to use
+    // their service account. In practice, though, I don't think this will be an issue.
```
jtk54: If we ever go to a multi-tenant Spinnaker, this may be a problem. We'll cross that bridge if/when we get to it.

jtk54:
LGTM

fiat-core/src/main/java/com/netflix/spinnaker/fiat/providers/DefaultServiceAccountProvider.java
ttomsu: +1

PR #19, opened 2016-07-22
Adds special handling for anonymous user. Periodically updates anonym…

ttomsu:
…ous and service accounts.

@jtk54 @duftler @cfieber PTAL.

@ajordens FYI

### DIFF
--- a/fiat-core/src/main/java/com/netflix/spinnaker/fiat/providers/DefaultServiceAccountProvider.java
+++ b/fiat-core/src/main/java/com/netflix/spinnaker/fiat/providers/DefaultServiceAccountProvider.java
@@ -74,10 +74,9 @@ public class DefaultServiceAccountProvider implements ServiceAccountProvider {
   public void updateServiceAccounts() {
     front50Service
         .getAllServiceAccounts()
-        .stream()
-        .forEach(serviceAccount -> {
+        .forEach(serviceAccount ->
           permissionsResolver.resolve(serviceAccount.getName())
-              .ifPresent(permissionsRepo::put);
-        });
+              .ifPresent(permissionsRepo::put)
+        );
   }
 }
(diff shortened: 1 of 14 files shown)


fiat-core/src/main/java/com/netflix/spinnaker/fiat/providers/DefaultServiceAccountProvider.java
```diff
@@ -83,4 +69,15 @@ public void afterPropertiesSet() {
         .map(serviceAccountsByName::get)
         .collect(Collectors.toSet());
   }
+
+  @Scheduled(initialDelay = 0, fixedDelay = 600000L)
+  public void updateServiceAccounts() {
+    front50Service
+        .getAllServiceAccounts()
+        .stream()
```
cfieber: minor but I am pretty sure you can `.forEach` without doing `.stream`

fiat-roles/src/main/groovy/com/netflix/spinnaker/fiat/permissions/DefaultPermissionsResolver.java
```diff
@@ -57,27 +57,52 @@
...
+  private boolean anonymousEnabled;
+
+  @Override
+  public Optional<UserPermission> resolveAnonymous() {
+    if (!anonymousEnabled) {
+      return Optional.empty();
+    }
+
+    val groups = new ArrayList<String>(0);
+    val accounts = accountProvider.getAccounts(groups);
```
cfieber: sorry because I don't have the whole project open at this moment - is accountProvider async loading all accounts and this call just operating off the cached view?

(wondering about the retry story if we are reaching directly out to other services here)

cfieber:
LGTM

fiat-roles/src/main/groovy/com/netflix/spinnaker/fiat/permissions/DefaultPermissionsResolver.java
ttomsu: No - it's hitting clouddriver every time. I'll definitely have to address the retry issue.

fiat-core/src/main/java/com/netflix/spinnaker/fiat/providers/DefaultServiceAccountProvider.java
ttomsu: Done.

PR #21, opened 2016-07-27
Adds retrying interceptor for all Retrofit calls. Introduces Provider…

ttomsu:
…Exception as common exception between all providers. 

Also introduces the notion of a "partial permission", where one or both of the dependent components were down at the time the permission was resolved. In this case, whatever is in the repo is not overrwritten. I'll take the next PR to make the server unhealthy if after some time or X failed calls, as the data is now stale.

@jtk54 @duftler @cfieber @ajordens - PTAL

### DIFF
--- a/fiat-web/config/fiat.yml
+++ b/fiat-web/config/fiat.yml
@@ -4,10 +4,9 @@ server:
 redis:
   connection: ${services.redis.connection:redis://localhost:6379}
 
-services:
-  connection:
-    retryBackoff:
-      maxSeconds: 5
+okHttpClient:
+  retries:
+    maxElapsedBackoffMs: 5000
 
 auth:
   anonymous:
--- a/fiat-web/config/fiat.yml
+++ b/fiat-web/config/fiat.yml
@@ -4,6 +4,11 @@ server:
 redis:
   connection: ${services.redis.connection:redis://localhost:6379}
 
+services:
+  connection:
+    retryBackoff:
+      maxSeconds: 5
+
 auth:
   anonymous:
     enabled: false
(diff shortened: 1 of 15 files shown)


fiat-web/src/main/java/com/netflix/spinnaker/fiat/config/FiatConfig.java
```diff
@@ -84,6 +96,43 @@ OkClient okClient() {
...
   }
+
+  @Slf4j
+  @AllArgsConstructor
+  private static class RetryingInterceptor implements Interceptor {
+
+    private long maxElapsedBackoffSeconds;
+
+    @Override
+    public Response intercept(Chain chain) throws IOException {
```
ajordens: Seems reasonable to me (it's only GET calls right from `fiat`?)

This is likely something that we'll want to have the usual metrics around (# of retries).

fiat-web/config/fiat.yml
```diff
@@ -4,6 +4,11 @@ server:
 redis:
   connection: ${services.redis.connection:redis://localhost:6379}
 
+services:
+  connection:
+    retryBackoff:
+      maxSeconds: 5
```
ajordens: Should this live in the normal `okHttpClient` config block vs. introducing a new section?

```
okHttpClient:
  retryBackOff:
    maxSeconds: 5
```

Consistency might see this as `okHttpClient.retries.maxElapsedBackoffMs` as we've used milliseconds for other config values.

Otherwise LGTM.

jtk54:
LGTM

fiat-web/src/main/java/com/netflix/spinnaker/fiat/config/FiatConfig.java
ttomsu: As of now only GETs, but I added a bit to ensure if someone ever POSTs or PATCHs to not retry.

fiat-web/config/fiat.yml
ttomsu: Done.

PR #22, opened 2016-07-28
Adds health indicator that won't signal the server is healthy until a…

ttomsu:
… successful service account sync.

If Fiat turns up before Front50, it will not become healthy (as indicated by `/health`) and keep retrying every 10s until a successful call to Front50 is made, at which point, the @Scheduler will resync every 10 minutes. If any periodic service account sync fails, it will continue to retry every 10s until a successful call. These retries are _independent_ of the Retrofit HTTP-level calls.  

If any period sync for user permissions fails 5 times (because either clouddriver or front50 are not responding), the Fiat service will go unhealthy. 

@jtk54 @duftler @cfieber @ajordens PTAL.

(FYI - I'll be out for the next week. Feel free to merge before then.)

### DIFF
--- a/fiat-core/src/main/java/com/netflix/spinnaker/fiat/providers/BaseProvider.java
+++ b/fiat-core/src/main/java/com/netflix/spinnaker/fiat/providers/BaseProvider.java
@@ -1,0 +1,59 @@
+/*
+ * Copyright 2016 Google, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.fiat.providers;
+
+import lombok.Data;
+import lombok.Setter;
+import org.springframework.beans.factory.annotation.Value;
+
+import java.util.concurrent.atomic.AtomicInteger;
+
+public class BaseProvider {
+
+  @Value("${unhealthy.threshold:5}")
+  @Setter
+  private int unhealthyThreshold;
+
+  private AtomicInteger failureCountSinceLastSuccess = new AtomicInteger(-1);
+
+  void failure() {
+    // Increment the failure count only if there has been at least 1 success() call. Otherwise,
+    // leave the default, which is considered unhealthy.
+    if (!failureCountSinceLastSuccess.compareAndSet(-1, -1)) {
+      failureCountSinceLastSuccess.incrementAndGet();
+    }
+  }
+
+  void success() {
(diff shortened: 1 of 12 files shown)


ttomsu:
Ping.

fiat-core/src/main/java/com/netflix/spinnaker/fiat/providers/BaseProvider.java
```diff
@@ -0,0 +1,59 @@
...
+  @Value("${unhealthy.threshold:5}")
+  @Setter
+  private int unhealthyThreshold;
+
+  private AtomicInteger failureCountSinceLastSuccess = new AtomicInteger(-1);
+
+  void failure() {
+    // Increment the failure count only if there has been at least 1 success() call. Otherwise,
+    // leave the default, which is considered unhealthy.
+    if (!failureCountSinceLastSuccess.compareAndSet(-1, -1)) {
```
jtk54: Are there any calls to providers that aren't on schedulers? If so, I think some exponential backoff (i.e. `sleep(2 ** failureCountSinceLastSuccess)`) would be useful just in case something takes a while to start up.

fiat-core/src/main/java/com/netflix/spinnaker/fiat/providers/BaseProvider.java
```diff
@@ -0,0 +1,59 @@
...
+
+package com.netflix.spinnaker.fiat.providers;
+
+import lombok.Data;
+import lombok.Setter;
+import org.springframework.beans.factory.annotation.Value;
+
+import java.util.concurrent.atomic.AtomicInteger;
+
+public class BaseProvider {
```
jtk54: Components like Gate and Orca could use this as well, maybe we can factor this out and include it there at some point as well?

fiat-core/src/main/java/com/netflix/spinnaker/fiat/providers/BaseProvider.java
jtk54: This would useful be in case where we make a burst of calls on startup to initialize.

ttomsu: A retrying interceptor with exponential backoff (up to 3 retries) was introduced in https://github.com/spinnaker/fiat/pull/21 and reviewed by... you :stuck_out_tongue_winking_eye:

fiat-core/src/main/java/com/netflix/spinnaker/fiat/providers/BaseProvider.java
ttomsu: The idea sounds like it has potential. Though I think we should come up with an easier way to share libraries/infrastructure across components. The kork version dance is a huge PITA and a deterrent to making a "common library".

fiat-core/src/main/java/com/netflix/spinnaker/fiat/providers/BaseProvider.java
jtk54: Heh.

fiat-core/src/main/java/com/netflix/spinnaker/fiat/providers/BaseProvider.java
jtk54: Ok makes sense.

jtk54:
LGTM

PR #24, opened 2016-08-26
Introduction of fiat-api module.

ttomsu:
The `fiat-api` module is a shared library that can be used by all of the other Spinnaker microservices to check if a user is authorized to view a particular resource.

The module defines the necessary pieces for Fiat integration, including: 
* A `Filter` to pull the Spinnaker username and put it into the Spring Security context.
* A Spring Security `PermissionEvaluator` that is used by the `PreAuthorize` and `PostFilter` annotations that will go on the secured endpoints.  
* A Retrofit interface for (strongly typed!!) REST calls going to Fiat.
* A config files to make it all work out-of-the-box when the `fiat-api` library is included in the classpath.

This is a relatively big PR, but I've broken it down into common commits for easier reviewability. Notes on the non `fiat-api` commits:
1.) Nested, non-static View classes are *really* nice for serialization - but not very nice (as in impossible) for ***de***serialization. So I had to make them static. In the process I introduced a new base class (`BaseView`) and interface (`Authorizable`) that are used in future commits.

2.) Filled out the `/authorize` API surface with `applications` and, more importantly, a generic "does user X have access to [read|write] resource Y"

(Personal aside: the strongly typed Retrofit interface would probably be better served with a protocol buffer interface, but maybe I'm showing my Google bias there ;-)  

@duftler @jtk54 @cfieber @ajordens PTAL

### DIFF
--- a/fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatPermissionEvaluator.java
+++ b/fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatPermissionEvaluator.java
@@ -1,0 +1,86 @@
+/*
+ * Copyright 2016 Google, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.fiat.shared;
+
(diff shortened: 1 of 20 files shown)


fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatPermissionEvaluator.java
```diff
@@ -0,0 +1,80 @@
...
+@Slf4j
+public class FiatPermissionEvaluator implements PermissionEvaluator {
+
+  @Autowired
+  @Setter
+  private FiatService fiatService;
+
+  @Value("${services.fiat.enabled:false}")
+  @Setter
+  private String fiatEnabled;
```
jtk54: Thanks Java8. Still think it's worth injecting a `FiatConfig` or something similar here to avoid the `String` types.

ttomsu: Maybe if more properties are introduced, but I don't think it's worth the complexity for just 1.

jtk54:
LGTM

PR #27, opened 2016-09-01
Adds command to pull whole permission into security context for @Post…

ttomsu:
…Filter optimization. Now a UserPermission with 500 applications will only make 1 network call instead of 500.

Also adds `isAdmin()` endpoint for potential future use.

@duftler @jtk54 @cfieber PTAL.

### DIFF
--- a/fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatPermissionEvaluator.java
+++ b/fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatPermissionEvaluator.java
@@ -17,6 +17,8 @@
 package com.netflix.spinnaker.fiat.shared;
 
 import com.netflix.spinnaker.fiat.model.Authorization;
+import com.netflix.spinnaker.fiat.model.UserPermission;
+import com.netflix.spinnaker.fiat.model.resources.Authorizable;
 import com.netflix.spinnaker.fiat.model.resources.Resource;
 import lombok.Setter;
 import lombok.extern.slf4j.Slf4j;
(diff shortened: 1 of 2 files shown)


fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatPermissionEvaluator.java
```diff
@@ -83,4 +91,65 @@ private boolean isAuthorized(String username, Resource resource, String resource
...
+      case APPLICATION:
+        return containsAuth.apply(permission.getApplications());
+      default:
+        return false;
+    }
+  }
+
+  @SuppressWarnings("unused")
+  public boolean isAdmin() {
+    return true; // TODO(ttomsu): Chosen by fair dice roll. Guaranteed to be random.
```
duftler: What could go wrong?

duftler:
LGTM.

fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatPermissionEvaluator.java
jtk54: What will actually be used to determine admin privileges (in future)? Certain group membership?

jtk54:
LGTM

fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatPermissionEvaluator.java
ttomsu: I'd imagine so. Everything else is group based, so I'd see no reason why not.

jtk54: SGTM

Issue #29, opened 2016-09-01
Bulk getUserPermission impl


ttomsu:
Current impl will crush the resource provider when there are a lot of users.

https://github.com/ttomsu/fiat/blob/c1c23041618aec7430d94d6a5d882fc7177a8957/fiat-roles/src/main/groovy/com/netflix/spinnaker/fiat/permissions/DefaultPermissionsResolver.java#L100-L100

PR #44, opened 2016-09-02
Google resource hierarchy

ttomsu:
As discussed with @jtk54 

This will be helpful with https://github.com/spinnaker/fiat/issues/29


jtk54:
LGTM

PR #45, opened 2016-09-06
Removes forcePut in exchange for Optional UserPermission, which is em…

ttomsu:
…pty in case any resource provider is unavailable.

@jtk54 @duftler PTAL.

tag: #29

### DIFF
--- a/fiat-roles/src/main/groovy/com/netflix/spinnaker/fiat/permissions/DefaultPermissionsResolver.java
+++ b/fiat-roles/src/main/groovy/com/netflix/spinnaker/fiat/permissions/DefaultPermissionsResolver.java
@@ -18,10 +18,8 @@ package com.netflix.spinnaker.fiat.permissions;
 
 import com.netflix.spinnaker.fiat.config.AnonymousUserConfig;
 import com.netflix.spinnaker.fiat.model.UserPermission;
-import com.netflix.spinnaker.fiat.providers.AccountProvider;
-import com.netflix.spinnaker.fiat.providers.ApplicationProvider;
 import com.netflix.spinnaker.fiat.providers.ProviderException;
-import com.netflix.spinnaker.fiat.providers.ServiceAccountProvider;
+import com.netflix.spinnaker.fiat.providers.ResourceProvider;
 import com.netflix.spinnaker.fiat.roles.UserRolesProvider;
 import lombok.NoArgsConstructor;
 import lombok.NonNull;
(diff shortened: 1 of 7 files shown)


fiat-roles/src/main/groovy/com/netflix/spinnaker/fiat/permissions/DefaultPermissionsResolver.java
```diff
@@ -48,15 +46,7 @@
...
-  private AccountProvider accountProvider;
-
-  @Autowired
-  @Setter
-  private ApplicationProvider applicationProvider;
-
-  @Autowired
-  @Setter
-  private ServiceAccountProvider serviceAccountProvider;
+  private List<ResourceProvider> resourceProviders;
```
jtk54: Nice

fiat-roles/src/main/groovy/com/netflix/spinnaker/fiat/permissions/DefaultPermissionsResolver.java
```diff
@@ -67,8 +57,8 @@
     if (!anonymousEnabled) {
       return Optional.empty();
     }
-    return Optional.of(getUserPermission(AnonymousUserConfig.ANONYMOUS_USERNAME,
-                                         new ArrayList<>(0) /* groups */));
+    return getUserPermission(AnonymousUserConfig.ANONYMOUS_USERNAME,
+                             new ArrayList<>(0) /* groups */);
```
jtk54: Why is the `<>` not filled in? The groups will be `String`s, right?

fiat-roles/src/main/groovy/com/netflix/spinnaker/fiat/permissions/DefaultPermissionsResolver.java
```diff
@@ -103,32 +93,26 @@
...
+    for (ResourceProvider provider : resourceProviders) {
       try {
-        permission.setServiceAccounts(serviceAccountProvider.getAll(groups));
+        permission.addResources(provider.getAll(groups));
       } catch (ProviderException pe) {
-        permission.setPartialPermission(true);
+        log.warn("Can't resolve permission for '%s' due to error with '%s'",
+                 userId,
+                 provider.getClass().getSimpleName());
+        log.debug("Error resolving UserPermission", pe);
```
jtk54: I think at least the short message/reason would be worth logging in `warn`.

jtk54:
LGTM after comments.

fiat-roles/src/main/groovy/com/netflix/spinnaker/fiat/permissions/DefaultPermissionsResolver.java
ttomsu: It's a feature of Java 7 - the diamond operator is used by the compiler to infer the type.
 
http://www.javaworld.com/article/2074080/core-java/jdk-7--the-diamond-operator.html

fiat-roles/src/main/groovy/com/netflix/spinnaker/fiat/permissions/DefaultPermissionsResolver.java
ttomsu: Good call. Done.

PR #47, opened 2016-09-06
Reimplements bulk user resolve to be significantly more efficient for…

### LINKED ISSUE #29: Bulk getUserPermission impl

ttomsu:
… large user bases. Fixes #29

@jtk54 @duftler PTAL


jtk54:
LGTM

Issue #52, opened 2016-09-14
Make single "all-users" resource access list.


ttomsu:
When a new resource is created (this came up when creating a new application, so we'll use application in the example), and that resource has no access restrictions, each user needs to have this new resource added to their list of things they can access. Performance will degrade with a large number of users, or frequently creating resources.

The solution seems to be adding an "anonymous" or "all-users" list of resources that all users have access to, and merging that into the access-controlled resources a user can access. I had to do this with the bulk user role loading already, which worked out well: https://github.com/spinnaker/fiat/blob/master/fiat-roles/src/main/groovy/com/netflix/spinnaker/fiat/permissions/DefaultPermissionsResolver.java#L160

This work is part of https://github.com/spinnaker/fiat/issues/51

PR #54, opened 2016-09-14
Replaces optional anonymous user with non-opptional 'unrestricted' us…

ttomsu:
…er, which keeps a single list of all unrestricted resources. These are merged at request time with the user's list of restricted access.

This one is kind of a big PR. It primarily removes the inclusion of unrestricted resources (accounts, applications, etc) from each individual user's Redis resource entries.

Secondly, it consolidates some of the annoying Resource switch statements in the RedisPermissionRepo now that the Resources have a common superclass.

Thirdly, it makes some of the test explicitly match known return values, rather than only comparing a handful of properties (such as size of accounts returns, account[0].name == "acct1", etc)

@duftler, @jtk54, @cfieber @ajordens PTAL

Closes https://github.com/spinnaker/fiat/issues/52

### DIFF
--- a/fiat-roles/src/main/groovy/com/netflix/spinnaker/fiat/permissions/RedisPermissionsRepository.java
+++ b/fiat-roles/src/main/groovy/com/netflix/spinnaker/fiat/permissions/RedisPermissionsRepository.java
@@ -42,7 +42,6 @@ import java.util.HashMap;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
-import java.util.function.Consumer;
 import java.util.function.Function;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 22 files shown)


fiat-roles/src/main/groovy/com/netflix/spinnaker/fiat/permissions/RedisPermissionsRepository.java
```diff
@@ -80,40 +83,40 @@
...
-            permission.getApplications().stream().forEach(putInValuesMap);
-            break;
-          case SERVICE_ACCOUNT:
-            permission.getServiceAccounts().stream().forEach(putInValuesMap);
-            break;
-        }
+        });
+
+    try (Jedis jedis = jedisSource.getJedis()) {
+      Pipeline pipeline1 = jedis.pipelined();
```
jtk54: Use more descriptive names for the pipelines please.

jtk54: Or at least comment and explain what is happening in each pipeline if they're doing different types of operations.

ttomsu: Done.

Issue #50, opened 2016-09-13
Make OkClient ConditionalOnMissing


ttomsu:
NFLX has an internal module for Front50 that provides the OkClient bean, so when the `fiat-api` module got included, the two clashed in an epic battle. Nobody wins when code collides.

ajordens:
:)

ttomsu:
@ajordens - can you give any more information on how you guys layer in your own module? I'm having some issues reproducing the issue and want ensure it gets fixed.

ttomsu:
And the exception you're getting, too.

ttomsu:
Comments from Slack chat for the record:

> ajordens [10:11 PM]  
I'm back now ... basically we had another config class that was also emitting a prototype'd OkClient.

> IMO the fix is perhaps moving RetrofitConfig from `fiat-core` to `fiat-web`. I haven't looked super close at what else is in `fiat-core` ... but because it's a transitive dependency of `fiat-api`, it'd probably be nice if it were extremely minimal. In theory something like the PermissionsRepositories shouldn't be necessary for consumers of `fiat-api` right? Also Front50Service/ClouddriverService?

PR #55, opened 2016-09-15
Fiat-api pruning

ttomsu:
Moves all of the `c.n.s.fiat.permissions/` and `provider/` files into `fiat-roles` so that they're not pulled into the `fiat-api` distribution. This leaves only the model classes, which someone could argue naming it `fiat-model` instead of `fiat-core`.

Furthermore, this PR moves the `RetrofitConfig` back into the `fiat-web` module, and relies on the client component to provide the `OkClient` & supporting classes. Turns out NFLX was broken because of conflicting OkClient prototype beans.

@ajordens @cfieber - can you take this for a spin before merging? Fiat follows the same publishToMaven --> make Front50 pull from mavenLocal() pattern [from Kork development](http://www.spinnaker.io/docs/kork-library-dev). 

@duftler @jtk54 PTAL, also.

Closes https://github.com/spinnaker/fiat/issues/50


ttomsu:
@ajordens - Also, what specificially should I change in the fiat-api gradle file to do what you mentioned in the chat (documented in #50 )

ajordens:
I'm good with these changes, I'll give them a whirl this morning to make sure everything hangs together.

Thanks for getting this together @ttomsu!

PR #57, opened 2016-09-15
Moves ServiceAccounts under /resources package

ttomsu:
Since ServiceAccounts are now `GroupAccessControlled`, they're no long a kind of "side" resource, so lets keep it with the rest of the resources. 

@jtk54 @duftler PTAL


ttomsu:
@lwander

Issue #53, opened 2016-09-14
Reverse index roles -> users for new resource creation


ttomsu:
When a new application gets created with access restrictions, we must find all users that belong in each of the specified access groups, and grant them explicit permission to this newly created application.

A reverse index of roles -> users is the most efficient way to accomplish this.

This work is part of https://github.com/spinnaker/fiat/issues/51

Issue #41, opened 2016-09-01
Roles passed through SAML assertion are dropped on first sync 


ttomsu:
Reported by @jvasallo, @ndcampbell

ttomsu:
This issue currently needs to be fixed in Gate, and ported over to Fiat.

jvasallo:
Thanks @ttomsu for this, I was going to mention that I couldn't open up an issue against gate :( 

Here is some more information on our setup (let me know if there is any info I can add to it):

**Ports:**
apiPort: 7777
defaultPort: 8084

**AuthN Methods:**
x509 has been enabled but only used by client scripts/applications to make API Calls to gate
SAML has been enabled for UI access via Deck

**authZ Methods**
We are passing memberOf in the SAML assertion with a list of groups the user is apart of.
We also have configured Github Group Membership for API access

**Experience:**
We have requiredGroupMembership in clouddriver that has both names from Github Groups and names from SAML Assertion. API Access works as expected since it is validating against GitHub groups. When a user logs in via SAML, they are able to browse around and make proper changes based on their groups. After roughly 10 minutes or so (the UserRolesSync period), the user is prompted with a "you have been logged out" message, and they are able to re-auth and hop back on. After looking at the logs, it appeared as though users who have logged in via SAML were being checked against Github group membership only; since they didn't exist they were getting logged out

**Expected:**
Users that login via SAML are able to continue to browse without getting signed out due to their group memberships being checked against Github. In short, SAML users get checked against SAML groups. Github group auth users are checked against github groups.

ttomsu:
Since you have an external roles provider already established, you should just drop the SAML roles from being sent over. The [code](https://github.com/spinnaker/gate/blob/master/gate-web/src/main/groovy/com/netflix/spinnaker/gate/security/rolesprovider/UserRolesSyncer.groovy#L96) logs a user out if there is a difference in roles between syncs, so if your external provider (GitHub Teams) kept returning the same roles each sync, your users would be able to stay logged in up until the timeout. Would that work for you?

Since these two are independent auth/z features, I'm less inclined to say we should support both simultaneously.

/cc: @jtk54

ttomsu:
Spoke with @jvasallo on Slack. They're setup for users to have SAML roles passed in, but are trying to use X509 certs with email addresses that map to GitHub Teams accounts. The code linked above does in fact work as intended by logging out any user whose roles have changed between syncs. 

As I see it, there are actually two issues here:
[...]

PR #60, opened 2016-09-19
Adds new Roles resource, to capture the roles a user received as part…

ttomsu:
… of the login process, or through the UserRoleProvider.

This new `Roles` object replaces various `String` types when constructing new `UserPermission` objects. It's primary advantage is specifying where the role came from.

The second big item in this PR is the establishment of a reverse index of roles. So each user has username --> `Map<String /*role name*/, Role>`, so now will there be a role name --> `Set<String /* username*/`. This reverse index is needed when new, restricted applications are created, and each member of a specified set of roles needs to have this new application added to its whitelist.

@duftler @jtk54 @ajordens @cfieber PTAL

tags: https://github.com/spinnaker/fiat/issues/53, https://github.com/spinnaker/fiat/issues/41

### DIFF
--- a/fiat-core/src/main/java/com/netflix/spinnaker/fiat/model/UserPermission.java
+++ b/fiat-core/src/main/java/com/netflix/spinnaker/fiat/model/UserPermission.java
@@ -19,6 +19,7 @@ package com.netflix.spinnaker.fiat.model;
 import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.netflix.spinnaker.fiat.model.resources.*;
 import lombok.Data;
+import lombok.EqualsAndHashCode;
 import lombok.NoArgsConstructor;
 
 import java.util.Collection;
@@ -31,14 +32,11 @@ import java.util.stream.Collectors;
 @Data
 public class UserPermission implements Viewable {
   private String id;
+
   private Set<Account> accounts = new HashSet<>();
   private Set<Application> applications = new HashSet<>();
   private Set<ServiceAccount> serviceAccounts = new HashSet<>();
-
-  @JsonIgnore
-  public boolean isEmpty() {
-    return accounts.isEmpty() && applications.isEmpty();
-  }
+  private Set<Role> roles = new HashSet<>();
 
   public void addResource(Resource resource) {
     addResources(Collections.singleton(resource));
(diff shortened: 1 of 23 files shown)


fiat-core/src/main/java/com/netflix/spinnaker/fiat/model/UserPermission.java
```diff
@@ -86,13 +87,15 @@ public View getView() {
   }
 
   @Data
+  @EqualsAndHashCode(callSuper = false)
```
jtk54: Do we know what gets generated here for the `Set`s below? For lists, equality is based on a simple address check - is the same true for sets? If equals between sets is handled right, this is fine.

fiat-roles/src/main/java/com/netflix/spinnaker/fiat/roles/google/GoogleDirectoryUserRolesProvider.java
```diff
@@ -167,6 +174,10 @@ private Directory getDirectoryService() {
         .build();
   }
 
+  private static Role toRole(Group g) {
```
jtk54: You could factor `toRole` out into a utility function that also takes the `source` as a parameter as well rather than having one per provider.

fiat-core/src/main/java/com/netflix/spinnaker/fiat/model/UserPermission.java
ttomsu: I added these to get ride of the compile warnings. Since the base class is empty, it should have no effect. Equals between sets is defined in the javadocs as:


> boolean equals(Object o)
> Compares the specified object with this set for equality. Returns true if the specified object is also a set, the two sets have the same size, and every member of the specified set is contained in this set (or equivalently, every member of this set is contained in the specified set). This definition ensures that the equals method works properly across different implementations of the set interface.

jtk54:
LGTM

fiat-roles/src/main/java/com/netflix/spinnaker/fiat/roles/google/GoogleDirectoryUserRolesProvider.java
jtk54: Ok, handling dependencies is far worse than a bit of code duplication.

Issue #51, opened 2016-09-13
Application Creation workflow


ttomsu:
There are some hiccups when it comes to deciding who can create a new application configuration, though the concept can also be applied Spinnaker wide when a user tries to access a resource.

The current workflow looks like:

1. Deck POSTs a Task to Gate with the `createApplication` type - [Source](https://github.com/spinnaker/deck/blob/master/app/scripts/modules/core/application/service/applications.write.service.js#L31)
1. Gate passes task onto Orca
1. Orca finds the [CreateApplicationStage](https://github.com/ttomsu/orca/blob/c9961c5ccdcf798f960a95154be938b11884fc27/orca-applications/src/main/groovy/com/netflix/spinnaker/orca/applications/pipelines/CreateApplicationStage.groovy#L28-L28), which invokes the [UpsertApplicationTask](https://github.com/ttomsu/orca/blob/7ac7cc39fbe72c0a630b8a269501aaf935621601/orca-applications/src/main/groovy/com/netflix/spinnaker/orca/applications/tasks/UpsertApplicationTask.groovy#L30-L30)
1. UpsertApplicationTask queries Front50 to see if an application with that name already exists
  - if so, call Front50.update
  - else, call Front50.create
1. Update the Permission
 
The hiccup exists with the call to see if the application in question already exists. The endpoint to do this is newly guarded by a Fiat authorization annotation. 

Previous to the Fiat integration, Front50 would return a 404. Now, the `@PreAuthorize` annotation looks up to see if the user has access to the (brand new) application. Obviously it doesn't, because it's brand new.

So, the question becomes how should the Front50 endpoint behave - it used to return 404 (Not Found), and now a 403 (Forbidden). 

Always returning a 403 is the "safest" from a security perspective, because it doesn't leak the existence of an app. Put another way, a malicious attacker could not write a script to check if "aaa", "aab", "aac" applications exist simply by comparing what returns 403 vs 404.

However, there are numerous places in the code with this code:

```groovy
try {
  return retrofitService.get()
} catch (RetrofitError re) {
  if (re.response.status == 404) {
    return null
  }
  throw re
}
```

If we went with 403, each one of these would have to be updated to include an `if(code == 403)` response.

Alternatively, a Retrofit Interceptor could be added to catch all 403s and make them `null`. 

An argument could be made that we're already assumed to be within a protected and trusted network, and that returning 404 could be more beneficial to whatever client application is asking for data. 

[...]

PR #61, opened 2016-09-19
Updates RolesController to accept a set of roles to update.

ttomsu:
This change enables other services (such as Front50) to invoke an update on a subset of users when a new resource is created (such as an application). It uses the reverse index of roles created in https://github.com/spinnaker/fiat/pull/60 to make the operation fast. 

Also creates a Logger to match the output of the rest of the logging output.

@jtk54 @duftler @ajordens @cfieber PTAL

Closes https://github.com/spinnaker/fiat/issues/51 after cutting a release.

### DIFF
--- a/fiat-roles/src/main/java/com/netflix/spinnaker/fiat/permissions/PermissionsRepository.java
+++ b/fiat-roles/src/main/java/com/netflix/spinnaker/fiat/permissions/PermissionsRepository.java
@@ -52,9 +52,10 @@ public interface PermissionsRepository {
 
   /**
    * Gets all UserPermissions in the repository that has at least 1 of the specified roles, keyed
-   * by user ID. If anyRoles is null, returns the same result as getAllById() (which
-   * includes the unrestricted user). If anyRoles is empty, it returns a map with a single entry for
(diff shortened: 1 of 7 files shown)


fiat-roles/src/main/java/com/netflix/spinnaker/fiat/permissions/PermissionsRepository.java
```diff
@@ -50,6 +51,16 @@
   Map<String, UserPermission> getAllById();
 
   /**
+   * Gets all UserPermissions in the repository that has at least 1 of the specified roles, keyed
+   * by user ID. If anyRoles is null, returns the same result as getAllById() (which
+   * includes the unrestricted user). If anyRoles is empty, it returns a map with a single entry for
+   * the unrestricted user.
```
jtk54: Can you explain why this is? I.e. "If `anyRoles` is non-empty, then by definition the unrestricted user is excluded... etc".

ttomsu: Done.

PR #62, opened 2016-09-19
Removes InMemoryPermissionRepo because due to lack of maintenance.

ttomsu:
InMem was used as a proof of concept, and since the Redis repo impl has gotten much more complex, I don't want to maintain the same complex behavior in two places.

@duftler @jtk54 PTAL


jtk54:
LGTM

duftler:
LGTM2.

PR #68, opened 2016-09-24
Don't delete clouddriver when uninstalling fiat

lwander:
I was looking through the project structure for help starting Halyard, and realized this file here probably doesn't do what it's intended to do.

@ttomsu


ttomsu:
I'm blaming that one on Cam's copy/paste.

PR #71, opened 2016-09-26
Wraps up some fiat-api configs into a bean

ttomsu:
@jtk54 @duftler @gregturn PTAL

### DIFF
--- a/fiat-api/src/main/java/com/netflix/spinnaker/config/FiatAuthenticationConfig.java
+++ b/fiat-api/src/main/java/com/netflix/spinnaker/config/FiatAuthenticationConfig.java
@@ -26,7 +26,10 @@ import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnExpression;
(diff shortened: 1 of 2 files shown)


fiat-api/src/main/java/com/netflix/spinnaker/config/FiatAuthenticationConfig.java
```diff
@@ -58,14 +62,17 @@
...
   private OkClient okClient;
 
-  @Value("${services.fiat.baseUrl}")
-  @Setter
-  private String fiatEndpoint;
+  @Bean
+  @ConfigurationProperties("services.fiat")
+  public FiatConfigurationProperties fiatConfigurationProperties() {
+    return new FiatConfigurationProperties();
+  }
```
gregturn: Instead of defining the whole bean here, you can simply put this at the top..

```
@EnableConfigurationProperties(FiatConfigurationProperties)
```

Put @ConfigurationProperties("services.fiat") on the class definition for the config prop class.

jtk54:
LGTM

fiat-api/src/main/java/com/netflix/spinnaker/config/FiatAuthenticationConfig.java
ttomsu: Done.

gregturn:
Perfect

Issue #33, opened 2016-09-01
Multiple Fiat instances need Redis lock to avoid stepping on each other's toes


ttomsu:
Similar to how Clouddriver acquires Redis locks to coordinate updates across instances.

@lwander - do you have any ideas around this?

lwander:
Sure thing - what's the work model that needs to be made mutually exclusive?

ttomsu:
Any mutating operation in this file: https://github.com/ttomsu/fiat/blob/b30d1f1cf081cf137150c69c1427696f49be8da7/fiat-roles/src/main/groovy/com/netflix/spinnaker/fiat/permissions/RedisPermissionsRepository.java#L61-L61

ttomsu:
And thank you!

ttomsu:
Ping @lwander - Will you be able to get to this by the end of the week?

lwander:
Will do my best, no promises though

lwander:
@ttomsu and I talked this through offline, and our conclusion was that the best way to handle this for now is to have 1 write instance of fiat (doing the role updates & large sync operations), and have the rest serve  read-only calls (since we expect this to accept a lot of traffic).

PR #73, opened 2016-09-27
Adds writeMode to restrict the UserRoleSyncer and UserPermission muta…

ttomsu:
…ting operations. Renames a few config properties.

The intent is to add the ability to horizontally scale Fiat the same way we do clouddriver - by having a single "master/write" instance, and multiple slave/read instances, all hooked up to the same Redis cluster.

@jtk54 @duftler @lwander @cfieber @ajordens PTAL

Fixes https://github.com/spinnaker/fiat/issues/33


jtk54:
LGTM. Needs a little rebase.

PR #75, opened 2016-09-27
Splits out Github and Google Groups role providers into their own mod…

ttomsu:
…ules.

@jtk54 @duftler @ajordens PTAL

@gbougeard - I didn't change anything here, so I don't see why everything wouldn't still work. Can you boot it up and confirm?


jtk54:
LTGM

PR #87, opened 2016-09-30
Adds first Fiat system test that tests the web endpoints with an...

ttomsu:
in-memory Redis and stubbed out Front50 and Clouddriver endpoints.

Also fixes a small XSS issue by only returning users who are in the repo.

This system test is a good step towards ensuring seemingly innocuous changes don't have unintended consequences.

@jtk54 @duftler @lwander PTAL


jtk54:
LGTM

PR #92, opened 2016-10-03
Replaces various spots of Collection.empty[Map|List|Set], which occas…

ttomsu:
…ionally throw UnsupportedOperationExceptions

@duftler @jtk54 PTAL


cfieber:
seems generally surprising that the return values from all of those spots should be expected to be mutable

ttomsu:
They don't all need to be mutable, but I don't want to hit another annoying UnsupportedOperationException. I left a few instances in there that I _know_ can be immutable.

jtk54:
These are used internally to Fiat only, so I think the mutable/immutable thing doesn't matter much. This LGTM.

PR #94, opened 2016-10-04
Debug message tweaks and some code formatting before next release

ttomsu:
@duftler @jtk54 PTAL

### DIFF
--- a/fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatPermissionEvaluator.java
+++ b/fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatPermissionEvaluator.java
@@ -52,16 +52,23 @@ public class FiatPermissionEvaluator implements PermissionEvaluator {
   private String fiatEnabled;
 
   @Override
-  public boolean hasPermission(Authentication authentication, Object resource, Object authorization) {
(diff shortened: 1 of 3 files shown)


fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatPermissionEvaluator.java
```diff
@@ -118,13 +135,20 @@ public boolean storeWholePermission() {
...
       view = fiatService.getUserPermission(username);
     } catch (RetrofitError re) {
-      String message = String.format("Cannot get whole user permission for user %s", username);
-      log.debug(message);
-      log.trace(message, re);
+      String message = String.format("Cannot get whole user permission for user %s. Cause: %s",
+                                     username,
+                                     re.getMessage());
+      if (log.isDebugEnabled()) {
+        log.debug(message, re);
```
jtk54: Any particular reason for `trace -> debug` and `debug -> info`?

ttomsu: The ability to tell users to turn on debug logging and getting everything instead of debug for some, and trace for others.

PR #97, opened 2016-10-06
Adds Clouddriver as source of applications.

ttomsu:
Depends on https://github.com/spinnaker/clouddriver/pull/1115

Uses unrestricted getAllApplications endpoint from Clouddriver to operate on the _complete_ list of Spinnaker applications.

@jtk54 @duftler @ajordens @cfieber PTAL 
/cc @gbougeard

### DIFF
--- a/fiat-roles/src/main/java/com/netflix/spinnaker/fiat/providers/internal/ClouddriverService.java
+++ b/fiat-roles/src/main/java/com/netflix/spinnaker/fiat/providers/internal/ClouddriverService.java
@@ -17,6 +17,7 @@
 package com.netflix.spinnaker.fiat.providers.internal;
 
 import com.netflix.spinnaker.fiat.model.resources.Account;
+import com.netflix.spinnaker.fiat.model.resources.Application;
 import retrofit.http.GET;
 
 import java.util.List;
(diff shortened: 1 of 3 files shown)


fiat-roles/src/main/java/com/netflix/spinnaker/fiat/providers/internal/ClouddriverService.java
```diff
@@ -25,4 +26,7 @@
 
   @GET("/credentials")
   List<Account> getAccounts();
+
+  @GET("/applications?restricted=false&expand=false")
```
jtk54: I think you can use arguments and annotations to avoid hardcoding this into the endpoint. Maybe something like https://github.com/spinnaker/clouddriver/blob/master/clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/core/services/Front50Service.groovy#L41?

ttomsu: I don't have any reason (and don't forsee any need in the future) to call them with different parameters, otherwise I would.

PR #104, opened 2016-10-12
Rm component

ttomsu:
@jtk54 FYI. Current Gate .deb package is broken, so I'm going to submit this and cut a release


jtk54:
LGTM

PR #107, opened 2016-10-20
Added stackdriver.

ewiseblatt:
@ttomsu 
I added some standard logic for determining whether or not you need mavenLocal. This is the same logic that is now in every other project (the heuristic earlier this week).


ttomsu:
It appears tests aren't passing.

ttomsu:
I rebased on master and this is now ready. Submitting.

Issue #125, opened 2017-01-10
requiredGroupMembership is null in application config


ttomsu:
Caused by an old issue where the Java object wasn't given a default empty collection for `requiredGroupMembership`

Manifests with a stack trace like this:

```
java.lang.NullPointerException: null
    at com.netflix.spinnaker.fiat.providers.DefaultApplicationProvider.lambda$getAllUnrestricted$3(DefaultApplicationProvider.java:83)
    at java.util.stream.ReferencePipeline$2$1.accept(ReferencePipeline.java:174)
    at java.util.HashMap$KeySpliterator.forEachRemaining(HashMap.java:1548)
    at java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:481)
    at java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:471)
    at java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:708)
    at java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
    at java.util.stream.ReferencePipeline.collect(ReferencePipeline.java:499)
    at com.netflix.spinnaker.fiat.providers.DefaultApplicationProvider.getAllUnrestricted(DefaultApplicationProvider.java:84)
    at com.netflix.spinnaker.fiat.permissions.DefaultPermissionsResolver.getUserPermission(DefaultPermissionsResolver.java:95)
    at com.netflix.spinnaker.fiat.permissions.DefaultPermissionsResolver.resolveUnrestrictedUser(DefaultPermissionsResolver.java:63)
    at com.netflix.spinnaker.fiat.roles.UserRolesSyncer.updateUserPermissions(UserRolesSyncer.java:147)
    at com.netflix.spinnaker.fiat.roles.UserRolesSyncer.syncAndReturn(UserRolesSyncer.java:97)
    at com.netflix.spinnaker.fiat.roles.UserRolesSyncer.sync(UserRolesSyncer.java:73)
    at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
    at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
    at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
    at java.lang.reflect.Method.invoke(Method.java:498)
[...]

PR #127, opened 2017-01-10
Ensure requiredGroupMembership value is never null

ttomsu:
Fixes https://github.com/spinnaker/fiat/issues/125

PTAL @duftler @jtk54


ttomsu:
It's already fixed in the code to write new app configs out as an empty list. There was a time, though, that some users created app configs with `requiredGroupMembership=null`.

Issue #30, opened 2016-09-01
Add additional UserRoleProviders 


ttomsu:
I've been seeing various users in Slack saying they can't/won't use Google Apps for Work or Github Teams for role provider. Some have proposed adding LDAP as both an authentication and authorization provider (aka UserRoleProvider).

If you'd like to see this or any other kind of provider, please make yourself known by commenting on this issue.

veeresh1982:
@ttomsu : I am interested to know more about this development. Can you please let me know the updates and timeline on supporting fiat with LDAP for both authentication/authorization.

ttomsu:
I've been working on getting the rest of the infrastructure integrated with Fiat, but realize additional user role providers will be necessary. That being said, I'd welcome a contribution to Gate that uses Spring Security LDAP and or an LDAP-based `UserRoleProvider` :-)

chongkong:
I guess SAML group is currently working?

ttomsu:
@chongkong - yes, see http://www.spinnaker.io/v1.0/docs/securing-spinnaker

JohnTheodore:
We won't use gapps/github, though we would use ldap for sure!

TheDigitalNinja:
I would love to see LDAP support.

To answer @JohnTheodore I would like to be able to plug directly into AWS Simple AD

tony612:
I really want this feature. And, what if we don't have SAML or LDAP? We build a auth service by ourself in our company to login to our inner services and manage the access. Is this possible for our situation? Maybe a webhook can be implemented(like k8s http://kubernetes.io/docs/admin/authorization/)?

rafaelcaricio:
Would be nice to be possible to use OAuth2 user scopes for authorization.

ttomsu:
Any brave souls out there interested in testing the initial LDAP functionality? I don't know if or how much different the ActiveDirectory work would be, but would be curious to talk with AD users too.

mtweten:
I've recently been working to try out the LDAP functionality in gate/fiat (we have an AD setup), and I think the only thing we'd need to be able to support the structure we have is some additional fields to the LDAP configuration. 

Some explanation of our setup: The DNs for our users have a CN of the formatted full name, e.g. `CN=Smith,John,OU=Users,OU=<some office location>,OU=Offices`, and the user id is in the sAMAccountName attribute. Using the userDnPattern field as it exists now, users in our setup would need to login with that full CN (Smith,John) instead of their user id. 

[...]

PR #131, opened 2017-01-12
Adds initial LDAP UserRoleProvider

ttomsu:
This enables Fiat to connect to an LDAP server to search for groups.

I attempted to write a test with an embedded LDAP server, but aparently ApacheDS uses an older, incompatible version of SLF4J, and I couldn't get it working.

PTAL @duftler @jtk54 

cc @scotmoor @rguthriemsft @ajordens @cfieber 
tag https://github.com/spinnaker/fiat/issues/30

### DIFF
--- a/fiat-ldap/src/main/java/com/netflix/spinnaker/fiat/roles/ldap/LdapUserRolesProvider.java
+++ b/fiat-ldap/src/main/java/com/netflix/spinnaker/fiat/roles/ldap/LdapUserRolesProvider.java
@@ -1,0 +1,123 @@
+/*
+ * Copyright 2017 Google, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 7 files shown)


fiat-ldap/src/main/java/com/netflix/spinnaker/fiat/roles/ldap/LdapUserRolesProvider.java
```diff
@@ -0,0 +1,123 @@
...
+    Set<String> userRoles = ldapTemplate.searchForSingleAttributeValues(
+        configProps.getGroupSearchBase(),
+        configProps.getGroupSearchFilter(),
+        params,
+        configProps.getGroupRoleAttributes());
+
+    log.debug("Got roles for user " + userId + ": " + userRoles);
+    return userRoles.stream()
+                    .map(role -> new Role(role).setSource(Role.Source.LDAP))
+                    .collect(Collectors.toList());
```
jtk54: You have `Set` as the return value, but are using `toList()` in the `collect`. Intended?

ttomsu: Give it another look - the method is returning a `List`

jtk54: My eyes and whitespace -_-. SGTM.

PR #137, opened 2017-01-19
Refactors and cleans up dead code in GitHub module

ttomsu:
This PR breaks apart the long `loadRoles` method into smaller methods.

Additionally, it sends the personal access token over the wire in an HTTP header, rather than the `access_code` query param. While it's nice to have the _actual_ URL printed in the logs, we probably shouldn't be purposefully printing access tokens in logs.

To further make this case, a few users have copy/pasted logs into the chat room, inadvertently sharing their access token. 

@gbougeard @jtk54 @duftler PTAL

cc @scotmoor


fiat-github/src/main/java/com/netflix/spinnaker/fiat/config/GitHubConfig.java
```diff
@@ -79,13 +74,13 @@ public void log(String message) {
...
 
-    @Getter
     @Setter
     private String accessToken;
 
     @Override
     public void intercept(RequestFacade request) {
-      request.addQueryParam("access_token", accessToken);
+      String encoded = Base64Utils.encodeToString(accessToken.getBytes(StandardCharsets.UTF_8));
+      request.addHeader("Authorization", "Basic " + encoded);
```
duftler: Not sure about the value of encoded here.

ttomsu: Added comment - all of these seem to work:

```
curl -u token https://api.github.com/user
curl -u ttomsu:token https://api.github.com/user
curl -u duftler:token https://api.github.com/user
```

Leading me to believe GitHub is stripping the "user:" part of the "user:password" Basic Auth standard.

fiat-github/src/main/java/com/netflix/spinnaker/fiat/config/GitHubConfig.java
```diff
@@ -79,13 +74,16 @@ public void log(String message) {
...
     private String accessToken;
 
     @Override
     public void intercept(RequestFacade request) {
-      request.addQueryParam("access_token", accessToken);
+      // Although the encoded string is normally "user:password" for Basic auth, GitHub appears to
+      // only look at the value to the right of the colon, or the whole value if no colon exists.
+      // This is nice for us, because we don't necessarily have a username to pass along anyway.
+      String encoded = Base64Utils.encodeToString(accessToken.getBytes(StandardCharsets.UTF_8));
+      request.addHeader("Authorization", "Basic " + encoded);
```
gbougeard: According the github doc it may be wrong : https://developer.github.com/v3/#authentication You might replace `Basic`  by `token`

ttomsu: Nice spot. Fixed.

fiat-github/src/main/java/com/netflix/spinnaker/fiat/roles/github/GithubTeamsUserRolesProvider.java
```diff
@@ -44,139 +46,122 @@
...
   }
 
-  @Override
-  public void afterPropertiesSet() throws Exception {
-    Assert.state(gitHubProperties.getOrganization ()!= null, "Supply an organization");
+  private void handleNon404s(RetrofitError e) {
+    String msg = "";
+    if (e.getKind() == RetrofitError.Kind.NETWORK) {
+      msg = String.format("Could not find the server %s", gitHubProperties.getBaseUrl());
+    } else if (e.getResponse().getStatus() == 401) {
```
gbougeard: Maybe it would be possible to handle the 403 when going over the rate limit? https://developer.github.com/v3/#rate-limiting this could be helpful to display the Github error message for troubleshooting

fiat-github/src/main/java/com/netflix/spinnaker/fiat/roles/github/GithubTeamsUserRolesProvider.java
```diff
@@ -44,139 +46,122 @@
...
+        handleNon404s(e);
       }
     }
     return false;
   }
 
-  @Override
-  public void afterPropertiesSet() throws Exception {
-    Assert.state(gitHubProperties.getOrganization ()!= null, "Supply an organization");
+  private void handleNon404s(RetrofitError e) {
```
scotmoor: Have you considered making these HTTP error handling methods part of the base class so that other providers calling a rest API could leverage them?

ttomsu: Not particularly. In this case, it was more able trying to make what the code was doing more apparent by moving the exception handling to another method.

fiat-github/src/main/java/com/netflix/spinnaker/fiat/roles/github/GithubTeamsUserRolesProvider.java
ttomsu: Added

PR #143, opened 2017-02-02
Gradle project version bump + cloudbuild trigger

ttomsu:
@cfieber PTAL

cc @duftler @skim1420


skim1420:
Missed my (very small) window, but what do you think of auto tagging images with commit hash?

PR #146, opened 2017-02-03
Slims container by deleting compile-time requirements.

ttomsu:
This PR drops the generated docker container from 533MB to **270MB**, and if using Google Cloud Builder (via cloudbuilder.yaml), it shrinks even further to **113.6MB** (which is basically Alpine container + Java 8 JRE + Fiat binary & libs).

I still have to ensure the thing actually _runs_, but at least it's building successfully. Will tackle that piece tomorrow.

@cfieber @duftler @skim1420 @lwander

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,8 +6,9 @@ COPY . workdir/
 
 WORKDIR workdir
 
-RUN GRADLE_USER_HOME=cache ./gradlew buildDeb -x test
-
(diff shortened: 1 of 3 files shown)


ttomsu:
I added `bash` and confirmed the container starts up and runs in a K8 environment!

skim1420:
LGTM

ttomsu:
cc @dennismartensson, too.

PR #147, opened 2017-02-13
Add endpoint to fetch user's roles

PerGon:
Create an endpoint to only fetch a user's roles - this makes it possible to get only the roles for a user without having to fetch the whole information as in the endpoint `getUserPermission`. Slightly more performant.


ttomsu:
Thanks for hammering this out!

PR #150, opened 2017-02-21
Fix startup issue with FiatService

gregturn:
Because Spring Security autoconfiguration happens early, current way of autoconfiguring FiatService will fail in later versions of Spring Boot. Instead, create a ObjectMapper on the fly to support creation.

### DIFF
--- a/fiat-api/src/main/java/com/netflix/spinnaker/config/FiatAuthenticationConfig.java
+++ b/fiat-api/src/main/java/com/netflix/spinnaker/config/FiatAuthenticationConfig.java
@@ -56,18 +56,12 @@ public class FiatAuthenticationConfig {
   @Setter
   private RestAdapter.LogLevel retrofitLogLevel = RestAdapter.LogLevel.BASIC;
 
-  @Autowired
-  @Setter
-  private ObjectMapper objectMapper;
-
-  @Autowired
-  @Setter
-  private OkClient okClient;
-
   @Bean
   @ConditionalOnMissingBean(FiatService.class) // Allows for override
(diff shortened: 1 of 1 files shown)


gregturn:
@spinnaker/reviewers PTAL

fiat-api/src/main/java/com/netflix/spinnaker/config/FiatAuthenticationConfig.java
```diff
@@ -38,6 +29,16 @@
 import org.springframework.security.config.annotation.web.builders.HttpSecurity;
 import org.springframework.security.config.annotation.web.configuration.EnableWebSecurity;
 import org.springframework.security.config.annotation.web.configuration.WebSecurityConfigurerAdapter;
+
+import com.fasterxml.jackson.databind.DeserializationFeature;
```
ttomsu: Any reason to split the imports here? For consistence, I'd rather them be kept together

gregturn:
I think it's my IDE. I have a Netflix format profile. I'll tweak and update this patch

gregturn:
@ttomsu I fixed the imports. I have an IDE Netflix profile that puts all com.netflix.* packages first. If that's not the case, I can adjust that as well.

duftler:
Looks like other files in this repo just sort them alphabetically: https://github.com/spinnaker/fiat/blob/master/fiat-core/src/main/java/com/netflix/spinnaker/fiat/model/UserPermission.java#L19-L20

gregturn:
@duftler Okay, I have a new IDE profile, where it doesn't do that. Pushed the polished update.

PR #154, opened 2017-03-09
fix(fiat): Remove services.fiat.autoconfig property in favor of @Enab…

ttomsu:
…leFiatAutoConfig annotation

This PR does two things:

1. Moves `@Configuration` files in `fiat-api` module outside of `com.netflix.spinnaker.config`, meaning they won't be automatically picked up by a component scan. References to `FiatClientConfigurationProperties` will need to be updated.
1. Each Spinnaker service now must "opt-in" to Fiat auto-config.

The second point is intentional because Fiat auto config should not be controlled by the user at runtime - it's determined by the code of the service, i.e. whether or not it provides it's own Spring Security config. The only service, to date, that does such a thing is Gate, which will not 

@cfieber @jtk54 PTAL

@gbougeard @gardleopard FYI

### DIFF
--- a/fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/EnableFiatAutoConfig.java
+++ b/fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/EnableFiatAutoConfig.java
@@ -1,0 +1,46 @@
+/*
+ * Copyright 2017 Google, Inc.
+ *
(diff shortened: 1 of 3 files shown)


ttomsu:
@jtk54 Correct. See https://github.com/spinnaker/front50/pull/213 as the first example

jtk54:
Got it, propagation delay.

cfieber:
:+1:

PR #159, opened 2017-03-29
feat(api): Adds shared cache of user permissions.

ttomsu:
This PR introduces a new cache of UserPermissions for users of the Fiat API module. It stores the whole permission for 20s (by default), even if the request could be serviced by a simple authorize call. This is the first step to scaling Fiat to a NFLX sized deployment.

This also means we no longer store the whole permission in the `Authentication` object.

@cfieber @jtk54 PTAL


cfieber:
I trust you
And @robzienert

PR #160, opened 2017-03-29
fix(service): disable content negotiation via URI path extension

cfieber:
fix a problem treating URIs that end with `.com` as a request for content-type
`application/ms-download`

turns out some people have email addresses that end with `.com` [1]

[1] - @cfieber, just now


ttomsu:
> turns out some people have email addresses that end with .com

citation needed.

PR #169, opened 2017-04-03
feat(authZ/ldap): Adds userSearchBase and userSearchFilter properties for LDAP

mtweten:
Exposes additional spring-security fields for the LDAP configuration that allow resolving the user via search instead of a user DN template.

In response to #30, with more detailed additional discussion [here](https://github.com/spinnaker/fiat/issues/30#issuecomment-289835132). Also created a PR for gate which adds the same fields for authn (https://github.com/spinnaker/gate/pull/369).


ttomsu:
Thanks for the PR!

PR #170, opened 2017-04-04
refactor(rpc): Introduces a wrapper around Retrofit APIs.

cfieber:
A subsequent change will make use of the wrapper classes to provide some temporary caching/
circuit breakers/fallbacks around calls to clouddriver and front50


fiat-roles/src/main/java/com/netflix/spinnaker/fiat/providers/DefaultAccountProvider.java
```diff
@@ -34,9 +34,12 @@
 @Component
 public class DefaultAccountProvider extends BaseProvider implements AccountProvider {
 
+  private final ClouddriverService clouddriverService;
+
   @Autowired
-  @Setter
-  private ClouddriverService clouddriverService;
+  public DefaultAccountProvider(ClouddriverService clouddriverService) {
```
cfieber: converted setter to constructor injection where I found usages of the updated services

ttomsu: I'd suggest using Lombok's `@RequiredArgConstructor` to make this easier.

https://projectlombok.org/features/Constructor.html

Issue #56, opened 2016-09-15
List of requiredGroupMembership --> requiredRoles with authorizations specified


ttomsu:
To further support authorization, Clouddriver and Front50 will need to be more expressive in their restrictions. Currently it looks like:

```
requiredGroupMembership: [
  "groupA",
  "groupB",
]
```

To support an extensible framework of authorizations, it should look something more like: 

```
permissions: {
  "read": ["groupA"],
  "write": ["groupB"]
}
```

In the case above, "read" would imply read-only, while "write" implies read-write. The beauty of an extensible system is we could make that name change with backwards compatibility.

This may not make it into the 1.0 release, depending on how distracted @cfieber gets between now and the EOQ.

/cc: @spinnaker/google-reviewers @skim1420 @ajordens

ttomsu:
Ping. Any update on this @cfieber ?

rfarrjr:
Curious what the status is on this feature.  Would be willing to help as my company would like this feature.

avram:
Given that the PRs all merged and I'm now seeing deprecation warnings for `requiredGroupMembership` in logs, is this issue closed?

spinnaker-release:
Migrated to https://github.com/spinnaker/spinnaker/issues/2439

PR #171, opened 2017-04-05
refactor(core): Major refactor to support read-only permissions

ttomsu:
This was much more work than I expected.

Major changes include:

1. Accounts, Applications, and Service Accounts now each have two new objects:
    * `Resource.Permissions` object that contains its Authorization->Roles (see https://github.com/spinnaker/clouddriver/pull/1544 and https://github.com/spinnaker/front50/pull/219)
    * `Set<Authorization>` that contains the specific authorizations for a particular user.

1. Supports backwards compatible `requiredGroupMembership` specification, granting both `READ` and `WRITE` authorizations. When found, it spits out warning messages, prompting the user to start using the `permissions` method instead.

A little more about point 1: We optimize multi-user resolution by only making 1 call to Front50/Clouddriver for the application/accounts config. These `Account` and `Application` objects must be cloned at the time `Set<Authorization>` is resolved and set for each user. Otherwise,  there would be cross-`Authorization` contamination.

tag https://github.com/spinnaker/fiat/issues/56


fiat-core/src/main/java/com/netflix/spinnaker/fiat/model/resources/BaseAccessControlled.java
```diff
@@ -0,0 +1,51 @@
...
+
+package com.netflix.spinnaker.fiat.model.resources;
+
+import com.netflix.spinnaker.fiat.model.Authorization;
+import lombok.extern.slf4j.Slf4j;
+
+import java.util.List;
+
+@Slf4j
+abstract class BaseAccessControlled<R extends Resource.AccessControlled> implements Resource.AccessControlled<R> {
```
jtk54: What's the plan for cutting support for `requiredGroupMembership`? Would you just remove this class and have the classes that used to extend this just implement the `AccessControlled` interface?

fiat-core/fiat-core.gradle
```diff
@@ -7,10 +7,12 @@ dependencies {
   compileOnly spinnaker.dependency("retrofitJackson")
 
   compileOnly spinnaker.dependency("lombok")
+  compileOnly spinnaker.dependency("slf4jApi")
```
cfieber: seems like compile not compileOnly

fiat-core/src/main/java/com/netflix/spinnaker/fiat/model/resources/BaseAccessControlled.java
ttomsu: Yeah, pretty much. We can make a call at some point in the future as to when "enough time has passed"

fiat-core/fiat-core.gradle
ttomsu: Fixed.

ttomsu:
I've refactored (again) to make `Permissions` immutable, and consumable by by Spring's `ConfigurationProperties` and Jackson's `ObjectMapper`. This was not trivial - it introduces a `Permissions.Builder` class, which Spring can shove values into, and those values get sanitized and locked upon `build()`.

This means we can use a single object across Fiat, Front50, and Clouddriver.

I've also removed a bunch of the `copyWithoutAuthorization` stuff. 

A user's restricted accounts/applications are still stored together at resolution time. However, at request time, the applicable authorizations are granted. Visually this looks like:

Resolution time:
Put in repo this info:
Travis has _some_ sort of access to --> App1, AccountA, AccountB
and has roles --> Role1

Clouddriver - get Travis' permission!
App1.getView(Role1) --> READ
AccountA.getView(Role1) --> READ,WRITE
AccountB.getView(Role1) --> WRITE

Issue #164, opened 2017-03-30
Make Service Accounts easier to use


ttomsu:
Remove requirement to have the identity provider actually have a user with that name.

ttomsu:
I'm considering changing to something like this:

A service account gets defined in Front50 like so:
```
{
"name": "foo",
"memberOf": ["groupA", "groupB"]
}
```
In this case, in order to access `foo`, a user must be a member of _all_ groups specified under `memberOf`.

This does two important things:

1. Resources are resolved the same way users are (same as today, but without the need to create additional groups/GitHub teams). If `account1` grants `groupA` the `READ` permission, then `foo` will have the `READ` permission.  
1. Prevents an escalation of privileges, because the service account can be used by users with the same (or less) authorizations the user has. 

Thoughts?

venturaville:
I'm 👍 on having the membeOf definition moved into the config instead.

"a user must be a member of all groups" ---> so by user here, I presume you mean the user who sets the pipeline config?  if so that makes sense

so the core auth* modules would add the service account to each of the groups dynamically coming back from github teams, or would they need to be separate groups?

msloes:
I am likely missing something... But since our users currently need to be part of the service account "team" (in Github) which also needs to be included in the Group Membership list per application - Does this change it so the service account no longer needs to be in Group Membership for each app?

ttomsu:
@venturaville:

> "a user must be a member of all groups" ---> so by user here, I presume you mean the user who sets the pipeline config? if so that makes sense

Yes - . Restated, "a _human_ user must be a member of all groups in order to save a pipeline config with the service account configured as the 'run as user'"

> so the core auth* modules would add the service account to each of the groups dynamically coming back from github teams, or would they need to be separate groups?

Not sure what you're saying here. A github user/team would no longer need to be created to represent the service account. Now, a service account would need to be manually created by an admin (for the time being). The authorizations are resolved the same way a user's permissions are (which happens _after_ the user's roles are acquired).

@msloes 
See above - a separate github user/team would no longer be needed. Instead, you'd only have to create a service account through Front50 like you do now.

PerGon:
This looks great to me! Thanks 🙇

Issue #163, opened 2017-03-30
Cache Front50 and Clouddriver responses


ttomsu:
In case they're unavailable.

Should refresh continuously, regardless if a role provider is used.

venturaville:
some more details on this would be helpful, including logging when they become unavailable, as well as how long they might be cached for (or making it configurable).

ttomsu:
We also came up with the idea of including `lastRefreshedTime` in the `/health` info, and possibly in the spectator metrics (so it can be alerted on)

PR #172, opened 2017-04-11
feat(serviceAccounts): Simplifies service accounts.

ttomsu:
Also pulls out provider health into it's own class, and caches response from
all resource providers for 20s.

As discussed in https://github.com/spinnaker/fiat/issues/164, this PR simplifies services accounts to just:
```
{
name: foo,
memberOf: [bar, baz]
}
```

additional tags: https://github.com/spinnaker/fiat/issues/163

PTAL @jtk54 @cfieber

### DIFF
--- a/fiat-roles/src/main/java/com/netflix/spinnaker/fiat/providers/BaseProvider.java
+++ b/fiat-roles/src/main/java/com/netflix/spinnaker/fiat/providers/BaseProvider.java
@@ -16,64 +16,44 @@
 
 package com.netflix.spinnaker.fiat.providers;
 
+import com.google.common.cache.Cache;
+import com.google.common.cache.CacheBuilder;
+import com.google.common.collect.ImmutableSet;
+import com.google.common.util.concurrent.UncheckedExecutionException;
+import com.netflix.spinnaker.fiat.config.ProviderCacheConfig;
 import com.netflix.spinnaker.fiat.model.resources.Resource;
 import com.netflix.spinnaker.fiat.model.resources.Role;
-import lombok.Data;
+import lombok.Getter;
 import lombok.NonNull;
-import lombok.Setter;
-import lombok.val;
-import org.springframework.beans.factory.annotation.Value;
+import lombok.extern.slf4j.Slf4j;
+import org.springframework.beans.factory.annotation.Autowired;
 
 import java.util.Set;
-import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
 
-public abstract class BaseProvider<R extends Resource.AccessControlled> implements ResourceProvider<R> {
+@Slf4j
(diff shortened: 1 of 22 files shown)


fiat-roles/src/main/java/com/netflix/spinnaker/fiat/providers/BaseProvider.java
```diff
@@ -16,72 +16,86 @@
...
-
-  @Data
-  class HealthView {
-    boolean providerHealthy = BaseProvider.this.isProviderHealthy();
-    int failureCountSinceLastSuccess = BaseProvider.this.failureCountSinceLastSuccess.get();
-  }
+  @Getter
+  private ProviderHealthTracker healthTracker = new ProviderHealthTracker();
 
+  private Cache<Integer, Set<R>> cache = buildCache(20);
```
jtk54: 20 is hardcoded here, didn't you have a configurable constant for this?

fiat-roles/src/main/java/com/netflix/spinnaker/fiat/providers/DefaultAccountProvider.java
```diff
@@ -36,18 +36,19 @@
...
 
   @Override
-  public Set<Account> getAll() throws ProviderException {
+  protected Set<Account> loadAll() throws ProviderException {
     try {
       val returnVal = clouddriverService.getAccounts().stream().collect(Collectors.toSet());
       success();
       return returnVal;
-    } catch (RetrofitError re) {
+    } catch (Exception e) {
```
jtk54: This seems like a wide net to cast here - I thought this was specifically when provider services didn't start up properly/in the correct order?

fiat-roles/src/main/java/com/netflix/spinnaker/fiat/providers/BaseProvider.java
ttomsu: It's hardcoded here, with the optional override via `setProviderCacheConfig` below

fiat-roles/src/main/java/com/netflix/spinnaker/fiat/providers/DefaultAccountProvider.java
ttomsu: Any exception should register as a failure - it's then converted to a `ProviderException` in BaseProvider

fiat-roles/src/main/java/com/netflix/spinnaker/fiat/providers/BaseProvider.java
jtk54: SGTM

fiat-roles/src/main/java/com/netflix/spinnaker/fiat/providers/DefaultAccountProvider.java
jtk54: SGTM2

ttomsu:
Ping @cfieber

PR #194, opened 2017-07-19
fix(cats): default timeout on interval to avoid over polling

rfarrjr:
After upgrading to Spinnaker 1.1 we noticed github rate limiting being triggered.   After some investigation it was noted that fiat was attempting to re-sync roles approximately every 10s.

I believe this was introduced from https://github.com/spinnaker/fiat/commit/e775f0d8a26b8df1cc59474180844f6115eb9e48 which introduced the cats feature.

It was not apparent to me when first look but the timeout on the AgentIntervalProvider is used as the ttl for the initial runkey.  If the agent is unable to complete in 10s the key expires and the when attempting to setup the next run (10min by default) the key has expired already.   The ClusteredAgentScheduler then schedules the agent again ~10s later.

This change will set the timeout to the default value of 2x the interval.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -36,7 +36,7 @@ allprojects {
   apply plugin: 'groovy'
 
   ext {
-    spinnakerDependenciesVersion = project.hasProperty('spinnakerDependenciesVersion') ? project.property('spinnakerDependenciesVersion') : '0.109.2'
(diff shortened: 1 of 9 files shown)


rfarrjr:
Depends on https://github.com/spinnaker/clouddriver/pull/1895

rfarrjr:
ping @spinnaker/reviewers

rfarrjr:
Closing to rebase given upstream fix.

PR #200, opened 2017-10-12
fix(authz/github): Add cache to team membership check to prevent excessive http requests

FaHeymann:
While setting up spinnaker on kubernetes using github for authentication and authorization we ran into the problem that fiat made too many requests to the github api, eventually running into rate limit restrictions making the spinnaker installation useless at that point.

This PR adds a cache to the problematic request to fix the problem temporarily.
Moving forward the attempt should probably be to reduce the amount of calls in the first place so that no such cache is necessary anymore.

### DIFF
--- a/fiat-github/src/main/java/com/netflix/spinnaker/fiat/roles/github/GitHubProperties.java
+++ b/fiat-github/src/main/java/com/netflix/spinnaker/fiat/roles/github/GitHubProperties.java
@@ -28,4 +28,6 @@ public class GitHubProperties {
   @Max(100L)
   @Min(1L)
   Integer paginationValue = 100;
+  @NotNull
+  Integer membershipCacheTTLSeconds = 60 * 1000;
 }
--- a/fiat-github/src/main/java/com/netflix/spinnaker/fiat/roles/github/GithubTeamsUserRolesProvider.java
+++ b/fiat-github/src/main/java/com/netflix/spinnaker/fiat/roles/github/GithubTeamsUserRolesProvider.java
@@ -75,6 +75,8 @@ public class GithubTeamsUserRolesProvider implements UserRolesProvider, Initiali
     Assert.state(gitHubProperties.getBaseUrl() != null, "Supply a base url");
   }
 
+  private static final String ACTIVE = "active";
+
   @Override
   public List<Role> loadRoles(String username) {
     log.debug("loadRoles for user " + username);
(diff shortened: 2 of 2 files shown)


fiat-github/src/main/java/com/netflix/spinnaker/fiat/roles/github/GithubTeamsUserRolesProvider.java
```diff
@@ -151,14 +160,33 @@ private boolean isMemberOfOrg(String username) {
   }
 
   private boolean isMemberOfTeam(Team t, String username) {
+    Integer CACHE_EXPIRES_AFTER_MINUTES = 60;
     String ACTIVE = "active";
+
+    if (this.teamMembershipCache == null) {
+      this.teamMembershipCache = CacheBuilder.newBuilder()
```
ttomsu: Can you break out this impl to another method?

fiat-github/src/main/java/com/netflix/spinnaker/fiat/roles/github/GithubTeamsUserRolesProvider.java
```diff
@@ -151,14 +160,33 @@ private boolean isMemberOfOrg(String username) {
   }
 
   private boolean isMemberOfTeam(Team t, String username) {
+    Integer CACHE_EXPIRES_AFTER_MINUTES = 60;
```
ttomsu: I'm not opposed to the defaults, but they should be configurable. Look at the GitHubProperties object for an example (it's injected into this class as a dependency already, which is nice).

ttomsu: the `ACTIVE` one should be moved to a private static member in the class, too.

fiat-github/src/main/java/com/netflix/spinnaker/fiat/roles/github/GithubTeamsUserRolesProvider.java
```diff
@@ -197,4 +225,9 @@ private static Role toRole(String name) {
 
     return emailGroupsMap;
   }
+
+  private @Data class CacheKey {
```
ttomsu: Annotations are generally put above the class declaration:

```
@Data
private class CacheKey {
}
```

ttomsu:
Thanks for the contribution!

fiat-github/src/main/java/com/netflix/spinnaker/fiat/roles/github/GitHubProperties.java
```diff
@@ -28,4 +28,6 @@
   @Max(100L)
   @Min(1L)
   Integer paginationValue = 100;
+  @NotNull
+  Integer membershipCacheTTLSeconds = 60 * 1000;
```
FaHeymann: @ttomsu Thanks for the review! Is this enough to make it configurable? Also should this be a parameter in halyard then?

ttomsu: Yeah, we'll need to plumb it through Halyard as well if users want to twiddle the value.

fiat-github/src/main/java/com/netflix/spinnaker/fiat/roles/github/GithubTeamsUserRolesProvider.java
```diff
@@ -151,18 +162,38 @@ private boolean isMemberOfOrg(String username) {
   }
 
   private boolean isMemberOfTeam(Team t, String username) {
-    String ACTIVE = "active";
+    if (this.teamMembershipCache == null) {
+      initializeTeamMembershipCache();
```
ttomsu: I missed in my initial review that this is an `InitializingBean`, which means you can add the `initializeTeamMembershipCache` call to the end of `afterPropertiesSet` method and remove this block. Sorry for not catching that the first time through.

ttomsu:
👍

ttomsu:
Thanks again!

PR #202, opened 2017-10-31
Add backoff and rety to GCP directory listings.

johnrengelman:
Adding some backoff to user group queries.


jasondtgt:
👍

ttomsu:
Thanks for the contribution!

PR #203, opened 2017-11-02
fix(UserRoleSyncer) a poll could take longer than 10 seconds

kevinawoo:
### The issue:
The github agent takes longer when there's more users/teams added, so we're seeing it take 12 seconds to run the agent now.
However the lock is set to expire after 10 seconds. 
This will allow another agent to run again. If there's multiple agents, it'll round robin until there's no more remaining API rate limit calls left. Github allows 5000 calls / hour.

### Fix:
Let's wait 30 seconds by default after the poll interval before allowing another instance of fiat to be the poller.

### DIFF
--- a/fiat-roles/src/main/java/com/netflix/spinnaker/fiat/config/CatsSchedulerConfig.java
+++ b/fiat-roles/src/main/java/com/netflix/spinnaker/fiat/config/CatsSchedulerConfig.java
@@ -54,6 +54,9 @@ public class CatsSchedulerConfig {
   @Value("${fiat.writeMode.syncDelayMs:600000}")
   String syncDelayMs;
 
+  @Value("${fiat.writeMode.syncDelayTimeoutMs:30000}")
+  String syncDelayTimeoutMs;
+
   @Bean
   NodeIdentity nodeIdentity() {
     return new DefaultNodeIdentity();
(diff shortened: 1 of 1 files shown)


ttomsu:
LGTM

fiat-roles/src/main/java/com/netflix/spinnaker/fiat/config/CatsSchedulerConfig.java
```diff
@@ -54,14 +54,19 @@
...
   NodeIdentity nodeIdentity() {
     return new DefaultNodeIdentity();
   }
 
   @Bean
   AgentIntervalProvider agentIntervalProvider() {
-    return new DefaultAgentIntervalProvider(Long.parseLong(syncDelayMs), 10000);
+    Long pollInterval = Long.parseLong(syncDelayMs);
+    Long timeout = Long.parseLong(syncDelayTimeoutMs);
+    return new DefaultAgentIntervalProvider(pollInterval, pollInterval + timeout);
```
demobox: If the intention here is "Let's wait 30 seconds by default after the poll interval before allowing another instance of fiat to be the poller.", should this be `(pollInterval, timeout);` rather than `(pollInterval, pollInterval + timeout);`?

@danielpeach Thanks for spotting!

kevinawoo: @demobox Yep, the defaults **mostly** work for everyone. This allows you to add extra time to the default.

kevinawoo: @demobox Yep, the defaults **mostly** work for everyone. This allows you to add extra time to when a poll cycle takes longer. 

The timeout is a redis key lock expiration time. So if a poller is working really hard (longer than 10 seconds), it won't expire the lock until we're 100% sure a poller is dead.

i.e.
Poll every 10s.
Poll job takes longer (12 seconds)
Since the poll job takes longer than 10s, don't start another poller until 10s + 30s.

If the instance of that poller were to go down, it could take `10s + 30s = 40s` for a new one to come back up.

PR #205, opened 2017-11-14
Back off on 500's or 403 rate limits.

johnrengelman:
Google returns a 403 on rate limit exceeded and the default retry only occurs on server 500s. However, this results in users intermittently loosing their authorities/roles in Fiat as only a partial list can be obtained.


markus-silpala-tgt:
LGTM.

ttomsu:
LGTM

PR #211, opened 2017-12-06
feat(admin) Add admin functionality to fiat

PerGon:
A quick summary of how it works is:

In the fiat configuration file, it will be possible to set a list of groups that will have admin capabilities.
```
fiat:
  admin:
    roles: 
      - sre_team
      - security_team
```
This will be merged with the permissions in each permissioned object (applications, accounts and service users). Members belonging to a group that is configured as Admin, have no restrictions in Spinnaker - meaning that they can access and change any resources.

The above example will mean that `sre_team` and `security_team` will have access (to view and to modify) to everything in Spinnaker.

Feedback is highly appreciated, thanks!

### DIFF
--- a/fiat-core/src/main/java/com/netflix/spinnaker/fiat/model/resources/Account.java
+++ b/fiat-core/src/main/java/com/netflix/spinnaker/fiat/model/resources/Account.java
@@ -17,12 +17,14 @@
 package com.netflix.spinnaker.fiat.model.resources;
 
 import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.google.common.collect.Sets;
 import com.netflix.spinnaker.fiat.model.Authorization;
 import lombok.Data;
 import lombok.EqualsAndHashCode;
 import lombok.NoArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
 
+import java.util.HashSet;
 import java.util.Set;
 
 @Data
@@ -35,8 +37,8 @@ public class Account extends BaseAccessControlled implements Viewable {
   private Permissions permissions = Permissions.EMPTY;
 
   @JsonIgnore
-  public View getView(Set<Role> userRoles) {
-    return new View(this, userRoles);
+  public View getView(Set<Role> userRoles, boolean isAdmin) {
+    return new View(this, userRoles, false);
   }
 
   @Data
@@ -46,9 +48,13 @@ public class Account extends BaseAccessControlled implements Viewable {
     String name;
     Set<Authorization> authorizations;
 
-    public View(Account account, Set<Role> userRoles) {
+    public View(Account account, Set<Role> userRoles, boolean isAdmin) {
       this.name = account.name;
-      this.authorizations = account.permissions.getAuthorizations(userRoles);
+      if (isAdmin) {
+        this.authorizations = Sets.newHashSet(Authorization.READ, Authorization.WRITE);
+      } else {
+        this.authorizations = account.permissions.getAuthorizations(userRoles);
+      }
     }
   }
 }
--- a/fiat-roles/src/main/java/com/netflix/spinnaker/fiat/config/FiatAdminConfig.java
+++ b/fiat-roles/src/main/java/com/netflix/spinnaker/fiat/config/FiatAdminConfig.java
@@ -1,0 +1,47 @@
+/*
+ * Copyright 2017 Schibsted ASA.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.fiat.config;
+
+import lombok.Data;
+import org.springframework.beans.factory.InitializingBean;
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.context.annotation.Configuration;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.stream.Collectors;
+
+@Data
+@Configuration
+@ConfigurationProperties("fiat")
+public class FiatAdminConfig implements InitializingBean {
+
+    private AdminRoles admin = new AdminRoles();
+
+    @Override
+    public void afterPropertiesSet() throws Exception {
+        admin.roles = admin.roles.stream()
+            .map(String::toLowerCase)
+            .map(String::trim)
+            .collect(Collectors.toList());
+    }
+
+    @Data
+    public static class AdminRoles {
+        private List<String> roles = new ArrayList<>();
+    }
+}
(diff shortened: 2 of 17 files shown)


PerGon:
@ttomsu do you think this is heading in the right direction? Thanks

fiat-roles/src/test/groovy/com/netflix/spinnaker/fiat/permissions/DefaultPermissionsResolverSpec.groovy
```diff
@@ -146,13 +153,41 @@ class DefaultPermissionsResolverSpec extends Specification {
     result == expected
   }
 
+  def "user should be admin"() {
+    fiatAdminConfig.getAdmin().getReadRoles().add("delivery-team")
```
ttomsu: It looks like after this test run, the shared fiat admin config is modified, making all remaining tests after this running under a different scenario. Since you're just creating the shared one to avoid NPE, I'd suggest doing something like:

```
.setFiatAdminConfig(new FiatAdminConfig())
``` 
on line 88, and then creating your custom config within this test.

fiat-roles/src/test/groovy/com/netflix/spinnaker/fiat/providers/DefaultServiceAccountProviderSpec.groovy
```diff
@@ -44,24 +48,26 @@ class DefaultServiceAccountProviderSpec extends Specification {
...
 
     where:
-    input           || expected
-    []              || []
-    ["a"]           || [aAcct]
-    ["b"]           || []
-    ["c"]           || []
-    ["a", "b"]      || [aAcct, bAcct]
-    ["a", "b", "c"] || [aAcct, bAcct]
+    input           || fiatAdmin           || expected
```
ttomsu: In Spock test tables, you generally put a single pipe `|` between each input/output, and the double pipe `||` between to signal the difference between inputs and outputs.

For example:
```
input1 | input2 | input3 || output1 | output2
```

PerGon:
Sweet! Thanks for the input - will address your comments as soon as possible (will be in a workshop the next few days).

Regarding Admin Read-Only or Read/Write, the idea is that you can set some groups to have admin real-only permissions. This will allow such admins, to view everything in spinnaker but not allow them to modify/create resources. The rationale is something like this: imagine the security team wants to view all applications, check which base images are being used, inspect all clusters, etc. But while they can see everything, they should not be allowed to modify applications or clusters (for example).
While a Read-Write Admin can do everything.
What do you think? Maybe we should simplify this in a first iteration by using only Admin Yes/No and possible later on add the Read-Only/Read-Write permissions for Admin?

PerGon:
@ttomsu I've simplified the admin functionality by having admin as a boolean - either a user is an admin or not.
As I said in the previous comment, maybe we can consider adding read/write separation later on.

fiat-roles/src/main/java/com/netflix/spinnaker/fiat/permissions/DefaultPermissionsResolver.java
```diff
@@ -90,14 +95,28 @@ public UserPermission resolveAndMerge(@NonNull ExternalUser user) {
     return getUserPermission(user.getId(), combo);
   }
 
+  private List<String> trimAndLowerCase(List<String> rolesList) {
+    return rolesList.stream()
```
ttomsu: The better place to do this is probably in the `FiatAdminConfig.AdminRoles` class. There are two ways to do this but only 1 might work:

1.) Override the `setAdminRoles(List list)` method to sanitize the input. This would be the ideal case, but unfortunately when Spring injects the values, it doesn't use this method - it uses `List list = getAdminRoles(); list.add(...); list.add(...);` etc

2.) Make `FiatAdminConfig` implement `InitializationBean` and implement the sanitization in the `afterPropertiesSet` method. I don't know if this will work in practice, though. Let me know what you find out. We may just have to keep this method here and go with it.

fiat-roles/src/main/java/com/netflix/spinnaker/fiat/config/FiatAdminConfig.java
```diff
@@ -0,0 +1,21 @@
...
+@Data
+@Configuration
+@ConfigurationProperties("fiat")
+public class FiatAdminConfig {
+
+    private AdminRoles admin = new AdminRoles();
+
+    @Data
+    public static class AdminRoles {
+        private List<String> adminRoles = new ArrayList<>();
```
ttomsu: The context of the `adminRoles` field is in the `AdminRoles` class, so you could just name it `roles`. That way, your config would look something like:

```
fiat:
  admin:
    roles:
      - foo
      - bar
  baz: qux (etc)
```

PerGon:
Hi @ttomsu ,

I've addressed your comments and added the persistence part. I've had build and deployed this to our test environment and it's working - I did some quick tests before the holiday season and it was working well. I'll do some more throughout tests tho.

I'm not very happy with the `View` part of the `model/resources` package - feels like a hack, but couldn't figure a better way, any idea?

fiat-core/src/main/java/com/netflix/spinnaker/fiat/model/resources/Account.java
```diff
@@ -46,9 +48,13 @@ public View getView(Set<Role> userRoles) {
     String name;
     Set<Authorization> authorizations;
 
-    public View(Account account, Set<Role> userRoles) {
+    public View(Account account, Set<Role> userRoles, boolean isAdmin) {
       this.name = account.name;
-      this.authorizations = account.permissions.getAuthorizations(userRoles);
+      if (isAdmin) {
```
PerGon: This if/else in here feels somewhat wrong 🤔 any idea of a better place?

PerGon:
Hello @ttomsu 

Do you think you will be able to have a look at this? I can then squash/rebase.

fiat-core/src/main/java/com/netflix/spinnaker/fiat/model/resources/Account.java
ttomsu: Not off the top of my head. I agree that passing a flag around isn't great, but I don't have a better idea.

fiat-roles/src/main/java/com/netflix/spinnaker/fiat/permissions/RedisPermissionsRepository.java
```diff
@@ -129,6 +131,17 @@ public RedisPermissionsRepository put(@NonNull UserPermission permission) {
     return this;
   }
 
+  //I'm not satisfied with this.
+  //TODO: Find out a wait to fetch admin statuses in batches
+  private boolean isUserAdmin(String userID) {
+    try (Jedis jedis = jedisSource.getJedis()) {
+      Pipeline p = jedis.pipelined();
+      Response<String> admin = p.hget(userID, adminKey());
```
ttomsu: If you're just doing 1 `hget`, you don't need a jedis `Pipeline`. You should just be able to `jedis.hget(...)`

fiat-roles/src/main/java/com/netflix/spinnaker/fiat/permissions/RedisPermissionsRepository.java
```diff
@@ -108,6 +109,7 @@ public RedisPermissionsRepository put(@NonNull UserPermission permission) {
 
       String userId = permission.getId();
       insertNewValuesPipeline.sadd(allUsersKey(), userId);
+      insertNewValuesPipeline.hset(userId, adminKey(), String.valueOf(permission.isAdmin()));
```
cfieber: Instead of a hash with the key userId with just a single entry adminKey -> "true" or "false" how about using a set with the key adminKey that just contains the users that are admins

means just one key to deal with, it means that bulk operations on admins are a bit easier. the save user just becomes a sadd or srem and the test for user being admin is sismember all of which are O(1)

PerGon:
Hi! I've just rebased and fixed conflicts.

@cfieber any chance you can look at the persistence side and see if you are happy with it? :)

Thanks!

kuberkaul:
exactly what @darwin67 said !

PerGon:
Hi @darwin67 and @kuberkaul 
Yes, I'm gonna try to do that as soon as possible.

PerGon:
It's a start: https://github.com/spinnaker/spinnaker.github.io/pull/1221

Will work on halyard support when possible

PR #213, opened 2018-01-19
fix(roles): ensure UserRolesSyncer always refreshes anonymous user

cfieber:
if we lose the redis data the user roles syncer would no longer reload the anonymous
permissions.

also this caps the number of retries that the refresh will do to something that aligns with
the caching agents timeout value. The agent will get rescheduled after that point by the
scheduler.


ttomsu:
LGTM

PR #215, opened 2018-02-06
chore(docker): Upgrade base image

wheleph:
See spinnaker/spinnaker#2298

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM java:8
+FROM openjdk:8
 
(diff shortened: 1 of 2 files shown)


ttomsu:
@lwander, @cfieber - My guess is we want to keep all services on the same JVM, yes?

@wheleph - Is there a reason you're looking to upgrade this one in particular?

wheleph:
@ttomsu, please have a look at the issue linked in the description. There are PRs for the rest of the services.

PR #225, opened 2018-05-07
fix(api): Stop parsing application names in fiat using frigga.

dibyom:
I tested this on my local setup. Before this fix, application names with dashes in them would return a 403 from Gate. Post this change, they work fine.

Clouddriver, Orca, Gate, and Front50 all need to be updated to the new version of fiat-api for this fix to work.

spinnaker/spinnaker#2513


dibyom:
@ttomsu do you see this change causing any issues?

ttomsu:
LGTM

PR #227, opened 2018-05-11
fix(roles/file): Handle empty file-based roles.

stewchen:


fiat-file/src/main/java/com/netflix/spinnaker/fiat/roles/file/FileBasedUserRolesProvider.java
```diff
@@ -102,7 +102,7 @@
     List<Role> roles;
 
     public List<Role> getRoles() {
-      return roles.stream()
+      return roles == null ? Collections.emptyList() : roles.stream()
```
ttomsu: Simple style nit - I think removing the turnary is easier to understand:

```
if (roles == null) {
  return Collections.emptyList();
}
return roles.stream()...
```

stewchen: Done.

PR #228, opened 2018-05-24
fix(hystrix): prevent silent NPE after timeout

danielpeach:
If Hystrix times out the command, `cause` is `null`. Hystrix gobbles the NPE and just rethrows the timeout, so we lose the fallback behavior.


fiat-roles/src/main/java/com/netflix/spinnaker/fiat/providers/internal/ClouddriverService.java
```diff
@@ -72,7 +72,8 @@ public void afterPropertiesSet() throws Exception {
           return accountCache.get();
         },
         (Throwable cause) -> {
-          log.warn("Falling back to account cache. Cause: " + cause.getMessage());
+          String message = cause != null ? "Cause: " + cause.getMessage() : "";
```
duftler: Could pull these 2 lines out into a utility method.

PR #229, opened 2018-06-04
feat(serviceAccounts): Change group behaviour of service accounts

gardleopard:
The current logic of service accounts is that if the service account is member of group a and b, the user that wants to use the service account needs to be a member of group a AND b. This patch changes this AND to an OR.

The rationale behind this is that when we have applications that are managed by multiple teams, they should be able to use a service account in the trigger and edit the pipelines without requiring the one to edit the pipelines to be a member of all the teams involved.


ttomsu:
This opens up escalation of privilege attacks.

lwander:
Safe to close @ttomsu?

ttomsu:
Unfortunately I haven't been able to come up with a better alternative for @gardleopard, but as stated before, I don't think this PR can be merged, so I'm closing. Perhaps @dibyom has had more time to think about it?

PR #230, opened 2018-06-06
feat(web): Support defaulting to `__unrestricted_user__` when no permissions found

ajordens:
This PR handles the case where a user attempts to access a service
directly without going through `gate`.

It is a common operational requirement to be able to access internal
APIs directly.

- bump spinnaker-dependencies to suppress 404 stack traces
- emit a metric from `getPermission` vs a debug log message

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -36,7 +36,7 @@ allprojects {
   apply plugin: 'groovy'
 
   ext {
-    spinnakerDependenciesVersion = project.hasProperty('spinnakerDependenciesVersion') ? project.property('spinnakerDependenciesVersion') : '0.154.3'
+    spinnakerDependenciesVersion = project.hasProperty('spinnakerDependenciesVersion') ? project.property('spinnakerDependenciesVersion') : '0.155.1'
   }
 
   def checkLocalVersions = [spinnakerDependenciesVersion: spinnakerDependenciesVersion]
--- a/fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatPermissionEvaluator.java
+++ b/fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatPermissionEvaluator.java
@@ -1,3 +1,4 @@
+
 /*
  * Copyright 2016 Google, Inc.
  *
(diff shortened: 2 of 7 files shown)


build.gradle
```diff
@@ -36,7 +36,7 @@ allprojects {
   apply plugin: 'groovy'
 
   ext {
-    spinnakerDependenciesVersion = project.hasProperty('spinnakerDependenciesVersion') ? project.property('spinnakerDependenciesVersion') : '0.154.3'
+    spinnakerDependenciesVersion = project.hasProperty('spinnakerDependenciesVersion') ? project.property('spinnakerDependenciesVersion') : '0.155.1'
```
ajordens: this latest rev suppresses the noise associated with stack traces being logged from `NotFoundException`.

fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatPermissionEvaluator.java
```diff
@@ -124,61 +123,48 @@ private String getUsername(Authentication authentication) {
     return username;
   }
 
-  private boolean isAuthorized(String username,
```
ajordens: Appeared that this was not being used anywhere (unless some service is calling this via groovy?)

ajordens: Appeared that this was not being used anywhere.

fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatPermissionEvaluator.java
```diff
@@ -124,61 +123,48 @@ private String getUsername(Authentication authentication) {
...
+    AtomicBoolean cacheHit = new AtomicBoolean(true);
+    boolean successfulLookup = true;
+
     try {
-      AtomicBoolean cacheHit = new AtomicBoolean(true);
       view = permissionsCache.get(username, () -> {
         cacheHit.set(false);
         return AuthenticatedRequest.propagate(() -> fiatService.getUserPermission(username)).call();
       });
-      log.debug("Fiat permission cache hit: " + cacheHit.get());
```
ajordens: Noise? Replaced with a metric.

fiat-web/src/main/java/com/netflix/spinnaker/fiat/controllers/AuthorizeController.java
```diff
@@ -191,4 +193,31 @@ public void getUserAuthorization(@PathVariable String userId,
 
     response.setStatus(HttpServletResponse.SC_NOT_FOUND);
   }
+
+  private Optional<UserPermission> getUserPermissionOrDefault(String userId) {
```
ajordens: Does this seem reasonable?

Aiming to support the use case of one of us needing to talk to an underlying service directly (ie. `clouddriver`) and not go through `gate`.

fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatPermissionEvaluator.java
cfieber: these are the spel permission functions in the `@Pre/PostAuthorize` annotations and I'm not sure they get detected as used. This one might not be used but might need some greppy type searches to confirm

ttomsu: cfieber@ is right that some of these methods will show up as unused, but are actually used in SPEL expressions. I don't think this one is though, based on this Github Org-level search: https://github.com/search?q=org%3Aspinnaker+isAuthorized&type=Code

fiat-web/src/main/java/com/netflix/spinnaker/fiat/controllers/AuthorizeController.java
ttomsu: Have you tried adding `-H X-SPINNAKER-USER:username@org.com` to your curl command?

fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatPermissionEvaluator.java
```diff
@@ -124,61 +123,48 @@ private String getUsername(Authentication authentication) {
...
+          ee.getCause().getMessage()
+      );
       if (log.isDebugEnabled()) {
         log.debug(message, ee.getCause());
       } else {
         log.info(message);
       }
     }
+
+    registry.counter(
```
ttomsu: I like this approach!

fiat-web/src/main/java/com/netflix/spinnaker/fiat/controllers/AuthorizeController.java
cfieber: This seems like a good starting point, we may eventually tease out some use cases around ops or propagating admin level credentials out of this, but I'm good with starting here

ajordens: That's a good point ... I did not set the header as I was coming in through the browser.

PR #231, opened 2018-06-07
fix(authn): invalidatePermission from cache

cfieber:
During login we want to read a fresh view of the permission
from fiat to support populating allowed accounts headers.

Also some cleanup in FiatAuthenticationConfig for easier autoconfigging.

### DIFF
--- a/fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatAuthenticationConfig.java
+++ b/fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatAuthenticationConfig.java
@@ -18,6 +18,7 @@ package com.netflix.spinnaker.fiat.shared;
 
 import com.fasterxml.jackson.databind.DeserializationFeature;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import com.netflix.spinnaker.config.OkHttpClientConfiguration;
 import com.netflix.spinnaker.okhttp.SpinnakerRequestInterceptor;
 import lombok.Setter;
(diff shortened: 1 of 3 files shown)


fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatAuthenticationConfig.java
```diff
@@ -54,22 +56,19 @@
   @Setter
   private RestAdapter.LogLevel retrofitLogLevel = RestAdapter.LogLevel.BASIC;
 
-  @Autowired
-  SpinnakerRequestInterceptor spinnakerRequestInterceptor;
-
   @Bean
   @ConditionalOnMissingBean(FiatService.class) // Allows for override
```
cfieber: I suspect this Conditional was to work around GateConfig explicitly configuring the FiatService and FiatPermissionEvaluator beans. This change removes the need for that duplication

fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatPermissionEvaluator.java
```diff
@@ -123,6 +123,10 @@ private String getUsername(Authentication authentication) {
     return username;
   }
 
+  public void invalidatePermission(String username) {
```
cfieber: needed during the login/logout flow to ensure we get a current user object in the fiat permission evaluator

ajordens:
LGTM

PR #232, opened 2018-06-08
feat(api): Support running with `management.security.enabled: false`

ajordens:
This PR re-orders how the `FiatAuthenticationFilter` is inserted into
the http chain.

Previously when management security was disabled, the caller was
redirected to `/login`.

- introduce a metric to track how often the unregistered user fallback is hit
- support dynamically disabling the permission evaluator

### DIFF
--- a/fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatAuthenticationConfig.java
+++ b/fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatAuthenticationConfig.java
@@ -26,18 +26,17 @@ import lombok.val;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
-import org.springframework.boot.autoconfigure.condition.ConditionalOnExpression;
+import org.springframework.beans.factory.annotation.Value;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
 import org.springframework.boot.context.properties.EnableConfigurationProperties;
-import org.springframework.boot.web.servlet.FilterRegistrationBean;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.ComponentScan;
 import org.springframework.context.annotation.Configuration;
-import org.springframework.core.Ordered;
 import org.springframework.security.config.annotation.method.configuration.EnableGlobalMethodSecurity;
 import org.springframework.security.config.annotation.web.builders.HttpSecurity;
 import org.springframework.security.config.annotation.web.configuration.EnableWebSecurity;
 import org.springframework.security.config.annotation.web.configuration.WebSecurityConfigurerAdapter;
+import org.springframework.security.web.context.SecurityContextPersistenceFilter;
 import retrofit.Endpoints;
 import retrofit.RestAdapter;
 import retrofit.client.Client;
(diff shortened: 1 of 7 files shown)


fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatPermissionEvaluator.java
```diff
@@ -206,4 +210,8 @@ private boolean permissionContains(UserPermission.View permission,
   public boolean isAdmin() {
     return true; // TODO(ttomsu): Chosen by fair dice roll. Guaranteed to be random.
   }
+
+  private boolean isEnabled() {
+    return dynamicConfigService.isEnabled("fiat", configProps.isEnabled());
```
ajordens: did a hack verification that the evaluator would start to short-circuit as expected when fiat was disabled

fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatAuthenticationConfig.java
```diff
@@ -78,25 +77,36 @@ public FiatService fiatService(FiatClientConfigurationProperties fiatConfigurati
...
+         * `SecurityContextHolder` being overridden with a null value.
+         *
+         * The null value then causes the `AnonymousAuthenticationFilter` to inject an "anonymousUser" which when
+         * passed over the wire to fiat is promptly rejected.
+         *
+         * This behavior is triggered when `management.security.enabled` is `false`.
+         */
+        http
+            .csrf().disable()
+            .addFilterAfter(new FiatAuthenticationFilter(), SecurityContextPersistenceFilter.class);
```
ajordens: Following along through the filter chain, I noticed that `SecurityContextPersistenceFilter` and `AnonymousAuthenticationFilter` basically blew away what the `FiatAuthenticationFilter` was setting in the `SecurityContextHolder`.

It's possible these additional filters came along for the ride when `management.security.enabled` was `false`. 

I'd specifically like to run a service with fiat enabled and the actuator endpoints left open, previously this resulted in a redirect to `/login`.

Would plan on testing with `gate` once https://github.com/spinnaker/gate/pull/553 is merged.

fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatAuthenticationConfig.java
```diff
@@ -78,25 +77,36 @@ public FiatService fiatService(FiatClientConfigurationProperties fiatConfigurati
   }
 
   @Bean
-  FilterRegistrationBean fiatFilterRegistrationBean(FiatAuthenticationFilter filter) {
-    val frb = new FilterRegistrationBean(filter);
-    frb.setOrder(Ordered.HIGHEST_PRECEDENCE + 1);
-    frb.addUrlPatterns("/*");
-    return frb;
+  FiatWebSecurityConfigurerAdapter fiatSecurityConfig(@Value("${services.fiat.enabled:false}") boolean isFiatEnabled) {
```
cfieber: wondering about this as a @Value here and then down below as a dynamic property

I suspect if we want to go the dynamic route then we need a static spring component configuration that always sets up the beans a consistent way, and the feature flag just controls whether those no-op or actually do things

fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatAuthenticationConfig.java
ajordens: I _think_ it works with gate ... at least to the extent that I've tested (gate launches fine and I can see requests go from gate -> fiat with the correct user).

fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatAuthenticationConfig.java
ajordens: This particular change was to support being able to hit the actuated endpoints.

fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatAuthenticationConfig.java
ttomsu: Have you tested with auth turned off?

ajordens: In the sense of turning off saml, x509 and ssl ... yes.

No calls to fiat (as expected).

PR #233, opened 2018-06-12
feat(api): Support enabling fiat at runtime

ajordens:
This PR will always register the `FiatAuthenticationFilter` in the
authentication chain _but_ allow for it to be disabled at runtime
depending on the dynamic config service.

The `FiatAuthenticationFilter` has been modified to no-op when
fiat would have previously been disabled.

A metric is also emitted (`fiat.enabled` to track whether fiat is
enabled on a service/instance basis).


ajordens:
Tested in `clouddriver` with `services.fiat.enabled: true` and `services.fiat.enabled: false` with expected results in each.

The aim of this PR is to allow Netflix to dynamically enable fiat on a service that was originally launched with it disabled (we do this with a custom `DynamicConfigService` implementation -- the OSS default will remain unchanged).

PR #234, opened 2018-06-13
fix(core): Remove the circular dependency on `clouddriver`

ajordens:
This PR makes use of the locking support in `kork` and removes a
dependency on `clouddriver:cats-redis`.

Redis usage has migrated to the `RedisClientDelegate` which in theory
opens us up to using dynomite at Netflix.

BREAKING CHANGE

Introduces a new config `fiat.writeMode.syncFailureDelayMs` as a means
of controlling the lock delay on failure vs success of a sync. This was
previously managed by `redis.poll.errorIntervalSeconds`.


fiat-roles/src/main/java/com/netflix/spinnaker/fiat/config/LockConfig.java
```diff
@@ -0,0 +1,52 @@
...
+    return Clock.systemDefaultZone();
+  }
+
+  @Bean
+  LockManager redisLockManager(Clock clock,
+                               Registry registry,
+                               ObjectMapper mapper,
+                               RedisClientDelegate redisClientDelegate) {
+    return new RedisLockManager(
+        null, // will fall back to running node name
```
ajordens: copy and pasta from `igor` and `swabbie`

fiat-roles/src/main/java/com/netflix/spinnaker/fiat/config/LockConfig.java
```diff
@@ -0,0 +1,52 @@
...
+        clock,
+        registry,
+        mapper,
+        redisClientDelegate,
+        Optional.empty(),
```
ajordens: heartbeat and lease durations are left at their defaults ... I don't know that this has any particular affect ... @jeyrschabu?

fiat-roles/src/main/java/com/netflix/spinnaker/fiat/permissions/RedisPermissionsRepository.java
```diff
@@ -103,33 +101,31 @@ public RedisPermissionsRepository put(@NonNull UserPermission permission) {
           }
         });
 
-    try (Jedis jedis = jedisSource.getJedis()) {
-      Pipeline deleteOldValuesPipeline = jedis.pipelined();
```
ajordens: Forgive me ... was there a reason for two pipelines before?

fiat-roles/src/main/java/com/netflix/spinnaker/fiat/permissions/RedisPermissionsRepository.java
```diff
@@ -103,33 +101,31 @@ public RedisPermissionsRepository put(@NonNull UserPermission permission) {
...
-      Pipeline insertNewValuesPipeline = jedis.pipelined();
+    try {
+      redisClientDelegate.withMultiKeyPipeline(pipeline -> {
```
ajordens: @robzienert -- does my respective use of the `redisClientDelegate` in here look good?

fiat-roles/src/main/java/com/netflix/spinnaker/fiat/roles/UserRolesSyncer.java
```diff
@@ -49,41 +48,44 @@
...
+    this.serviceAccountProvider = serviceAccountProvider;
+    this.healthIndicator = healthIndicator;
+
+    this.retryIntervalMs = retryIntervalMs;
+    this.syncDelayMs = syncDelayMs;
+    this.syncDelayTimeoutMs = syncDelayTimeoutMs;
+  }
 
+  @Scheduled(fixedDelayString = "${fiat.writeMode.syncDelayMs:600000}")
+  public void schedule() {
```
ajordens: I need to circle around and actually have this obey discovery status for start/stop

fiat-roles/src/main/java/com/netflix/spinnaker/fiat/roles/UserRolesSyncer.java
```diff
@@ -49,41 +48,44 @@
...
+    LockManager.LockOptions lockOptions = new LockManager.LockOptions()
+        .withLockName("Fiat.UserRolesSyncer".toLowerCase())
+        .withMaximumLockDuration(Duration.ofMillis(syncDelayMs + syncDelayTimeoutMs));
```
ajordens: pretty sure this was what the old behavior in `CatsSchedulerConfig` was doing

ajordens:
Need to run through some manual testing, _at_ this point I've validated that fiat is able to sync.

fiat-roles/src/main/java/com/netflix/spinnaker/fiat/config/RedisConfig.java
```diff
@@ -48,6 +40,11 @@ public JedisPool jedisPool(@Value("${redis.connection:redis://localhost:6379}")
     return createPool(redisPoolConfig, connection, timeout);
   }
 
+  @Bean
+  RedisClientDelegate redisClientDelegate(JedisPool jedisPool) {
+    return new JedisClientDelegate(jedisPool);
+  }
```
robzienert: Not required, but raising your attention if you missed it: If you want all of the kork auto-configuration stuff (interchangeable redis/dynomite configs, named client selector, etc), you can also do: https://github.com/spinnaker/kork/blob/master/kork-jedis/src/main/java/com/netflix/spinnaker/kork/jedis/JedisClientConfiguration.java

fiat-roles/src/main/java/com/netflix/spinnaker/fiat/config/LockConfig.java
jeyrschabu: Default values are reasonable. The rule of thumb is to keep `heartbeatRateMillis` less than the lock ttl to avoid expiring locks. In Igor I had to watch metrics and tweak these values accordingly.

fiat-roles/src/main/java/com/netflix/spinnaker/fiat/roles/UserRolesSyncer.java
jeyrschabu: Are we locking on one key only?

cfieber:
There is some behaviour in the cats scheduler that specifically went in to support the usage in fiat - specifically the retry interval on success vs failure

Looking to see if we still have that with the new implementation

It was to prevent getting throttled on the group source, but I'd need to dig in a bit more for exactly how it was used in the original implementation

ajordens:
FYI @rfarrjr -- You had added some support in clouddriver for controlling the failure vs. success interval. 

This PR removes the dependency that `fiat` has on `clouddriver` and thus the config key has changed to `fiat.writeMode.syncFailureDelayMs`.

rfarrjr:
Pulling this in, but this looks like it also may have resolved https://github.com/spinnaker/spinnaker/issues/2450.   We will verify.

PR #236, opened 2018-06-18
fix(tests): Ensure that the `UserRolesSyncer` does not run mid-test

ajordens:


ajordens:
Observed that the `UserRolesSyncer` may mid-test (based on it's scheduled interval) and overwrite manual manipulations of the permissions repository that the integration tests may have made.

This was failing a test in travis -- I do plan on circling around and altering the `UserRolesSyncer` such that it'll only run when the service is UP (similar to elsewhere where we're running schedulers).

PR #237, opened 2018-06-19
feat(core): User role syncing should pause when not in discovery

ajordens:
If discovery is not available, the `UserRolesSyncer` will be enabled
and run as expected.

Follows a similar pattern used in other services.

### DIFF
--- a/fiat-roles/src/main/java/com/netflix/spinnaker/fiat/roles/UserRolesSyncer.java
+++ b/fiat-roles/src/main/java/com/netflix/spinnaker/fiat/roles/UserRolesSyncer.java
@@ -17,6 +17,8 @@
 package com.netflix.spinnaker.fiat.roles;
 
 import com.diffplug.common.base.Functions;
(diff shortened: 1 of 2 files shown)


fiat-roles/src/main/java/com/netflix/spinnaker/fiat/roles/UserRolesSyncer.java
```diff
@@ -197,4 +220,21 @@ public long updateUserPermissions(Map<String, UserPermission> permissionsById) {
     log.info("Synced {} non-anonymous user roles.", count);
     return count;
   }
+
+  private boolean isInService() {
+    InstanceInfo.InstanceStatus remoteStatus = null;
+    if (discoveryClient.isPresent()) {
```
cfieber: could do `discoveryClient.map(DiscoveryClient::getInstanceRemoteStatus).orElse(InstanceInfo.InstanceStatus.UP)`

PR #240, opened 2018-06-27
feat(core): Support a fallback to legacy account permissions

ajordens:
Should `fiat` ever be unavailable, this supports a fallback to a
`UserPermission` that is constructed based on the contents of
the `AuthenticatedRequest`.

In the event of a legacy fallback, access to _all_ applications will be
explicitly granted.

It is not on by default and must be explicited opt'd enabled via:

```
fiat:
  services:
    legacyFallback: true
```

or by setting the `fiat.legacyFallback.enabled` dynamic property.

### DIFF
--- a/fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatPermissionEvaluator.java
+++ b/fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatPermissionEvaluator.java
@@ -24,9 +24,9 @@ import com.netflix.spectator.api.Id;
 import com.netflix.spectator.api.Registry;
 import com.netflix.spinnaker.fiat.model.Authorization;
 import com.netflix.spinnaker.fiat.model.UserPermission;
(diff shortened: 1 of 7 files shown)


ajordens:
Warrants a test or two around the new `legacyFallback` behavior

fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatPermissionEvaluator.java
```diff
@@ -74,6 +78,7 @@ public FiatPermissionEvaluator(Registry registry,
         .build();
 
     this.getPermissionCounterId = registry.createId("fiat.getPermission");
+    this.legacyFallback = configProps.isLegacyFallback();
```
cfieber: worth adding a dynamic props feature flag to this maybe

ajordens: That's reasonable, I'll get that in.

PR #242, opened 2018-06-29
feat(roles): Support passing additional user details to `UserRolesProvider`

ajordens:


ajordens:
Plan is to have a custom `UserRolesProvider` that can do something Netflix-specific when it comes to dealing with our legacy minted certificates.

It's helpful to know what roles were actually supplied hence the interface change.

ajordens:
Verified that we can have a custom `UserRolesProvider` that does what we need to do.

PR #244, opened 2018-07-09
fix(roles/sync): Schedule sync every fiat.writeMode.syncDelayMs

dibyom:
Currently, a sync is scheduled with a fixed delay of 30s while the userRoleSyncer lock expires after `syncDelay`(defaults to 10 min). This leads to a ` Could not acquire already taken lock Lock` log message every 30s.

This PR changes the role syncer to be scheduled every `syncDelay` ms.

@danielpeach @ajordens

### DIFF
--- a/fiat-roles/src/main/java/com/netflix/spinnaker/fiat/roles/UserRolesSyncer.java
+++ b/fiat-roles/src/main/java/com/netflix/spinnaker/fiat/roles/UserRolesSyncer.java
@@ -105,7 +105,7 @@ public class UserRolesSyncer implements ApplicationListener<RemoteStatusChangedE
     isEnabled.set(isInService());
   }
 
-  @Scheduled(fixedDelay = 30000L)
+  @Scheduled(fixedDelayString = "${fiat.writeMode.syncDelayMs?:600000}")
   public void schedule() {
     if (syncDelayMs < 0 || !isEnabled.get()) {
       return;


fiat-roles/src/main/java/com/netflix/spinnaker/fiat/roles/UserRolesSyncer.java
```diff
@@ -105,7 +105,7 @@ public void onApplicationEvent(RemoteStatusChangedEvent event) {
     isEnabled.set(isInService());
   }
 
-  @Scheduled(fixedDelay = 30000L)
+  @Scheduled(fixedDelayString = "${fiat.writeMode.syncDelayMs?:600000}")
```
danielpeach: Not sure if the `?` will work here - would change to `syncDelayMs:600000`. I basically agree with @ajordens here, since polling for the lock every 30 seconds is at worst doing no harm.

fiat-roles/src/main/java/com/netflix/spinnaker/fiat/roles/UserRolesSyncer.java
```diff
@@ -105,7 +106,7 @@ public void onApplicationEvent(RemoteStatusChangedEvent event) {
     isEnabled.set(isInService());
   }
 
-  @Scheduled(fixedDelay = 30000L)
+  @Scheduled(fixedDelayString = "#{userRolesSyncer.getSyncDelayMs()}")
```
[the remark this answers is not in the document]
dibyom: The log message comes from the Kork library. Was hesitant to change it there since the message might be useful for other services using the lock library.

Agreed that grabbing a lock more frequently than the delay would be ideal though I think the schedule should be calculated based on the syncDelay say every syncDelay/2 ms? What do you think?

This isn't a big issue at all but it just eases debugging Fiat since currently most of the logs are just filled with `could  not acquire lock messages`.

fiat-roles/src/main/java/com/netflix/spinnaker/fiat/roles/UserRolesSyncer.java
dibyom: The `?` is needed otherwise the `RolesControllerSpec` fails with:
```
Caused by: java.lang.IllegalArgumentException: null
	at java.util.concurrent.ScheduledThreadPoolExecutor.scheduleWithFixedDelay(ScheduledThreadPoolExecutor.java:611)
```

ttomsu: Silly Java not having the Elvis operator :-1:

ajordens:
I've dropped that kork log from `INFO` to `DEBUG`, we couldn't come up with a reason on our side to keep it as-is.

spinnaker/fiat#246 pulls it in.

With that change, I'm not sure we need to change the default behavior? (at the very least I probably wouldn't tie it to `fiat.writeMode.syncDelayMs` as I'd still like a means to configure it to a shorter interval).

dibyom:
I think just dropping it to `DEBUG` is sufficient. I'll close this PR. Thanks!

PR #247, opened 2018-07-11
fix(roles/github): Reduce burden of github reads for teams and make re…

rfarrjr:
…silient to rate-limiting.

As our github org has grown we have had more and more difficulty with rate-limits from github.   This change reworks the calls to github to cache the teams and user information with a default backing cache.

The cache can be tweaked with configuration:
```yaml
auth:
  groupMembership:
    github:
      membershipCacheTTLSeconds: 600
      membershipCacheTeamsSize: 1000
```

### DIFF
--- a/fiat-github/src/main/java/com/netflix/spinnaker/fiat/roles/github/GitHubProperties.java
+++ b/fiat-github/src/main/java/com/netflix/spinnaker/fiat/roles/github/GitHubProperties.java
@@ -29,5 +29,7 @@ public class GitHubProperties {
   @Min(1L)
   Integer paginationValue = 100;
   @NotNull
-  Integer membershipCacheTTLSeconds = 60 * 60; // 1 hour
+  Integer membershipCacheTTLSeconds = 60 * 10; // 10 min time to refresh
+  @NotNull
+  Integer membershipCacheTeamsSize = 1000; // 1000 github teams
 }
--- a/fiat-github/src/main/java/com/netflix/spinnaker/fiat/roles/github/GithubTeamsUserRolesProvider.java
+++ b/fiat-github/src/main/java/com/netflix/spinnaker/fiat/roles/github/GithubTeamsUserRolesProvider.java
@@ -82,7 +82,7 @@ public class GithubTeamsUserRolesProvider implements UserRolesProvider, Initiali
   private LoadingCache<String, Set<String>> membersCache;
 
   private LoadingCache<String, List<Team>> teamsCache;
-  
+
   private LoadingCache<Long, Set<String>> teamMembershipCache;
 
   private static final String ACTIVE = "active";
@@ -98,8 +98,9 @@ public class GithubTeamsUserRolesProvider implements UserRolesProvider, Initiali
   }
 
   private void initializeMembersCache() {
+    // Note if multiple github orgs is ever supported the maximumSize will need to change
     this.membersCache = CacheBuilder.newBuilder()
-        .maximumSize(1)
+        .maximumSize(1) //This will only be a cache of one entry keyed by org name.
         .refreshAfterWrite(this.gitHubProperties.getMembershipCacheTTLSeconds(), TimeUnit.SECONDS)
         .build(new CacheLoader<String, Set<String>>() {
           public Set<String> load(String key) {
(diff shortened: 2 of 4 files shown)


fiat-github/src/main/java/com/netflix/spinnaker/fiat/roles/github/GitHubProperties.java
```diff
@@ -29,5 +29,7 @@
   @Min(1L)
   Integer paginationValue = 100;
   @NotNull
-  Integer membershipCacheTTLSeconds = 60 * 60; // 1 hour
+  Integer membershipCacheTTLSeconds = 60 * 10; // 10 min time to refresh
```
ttomsu: Is there a reason to bring this TTL down? Wouldn't a lower TTL cause more requests?

fiat-github/src/main/java/com/netflix/spinnaker/fiat/roles/github/GithubTeamsUserRolesProvider.java
```diff
@@ -77,29 +92,114 @@ public void afterPropertiesSet() throws Exception {
...
     this.initializeTeamMembershipCache();
   }
 
-  private void initializeTeamMembershipCache() {
-    this.teamMembershipCache = CacheBuilder.newBuilder()
-        .maximumSize(1000)
-        .expireAfterWrite(this.gitHubProperties.getMembershipCacheTTLSeconds(), TimeUnit.SECONDS)
+  private void initializeMembersCache() {
+    this.membersCache = CacheBuilder.newBuilder()
+        .maximumSize(1)
```
ttomsu: Uh... max size of 1? If this intentional, I'd add a comment as to why.

ttomsu:
I like the change to using the reloading cache, but IIUC this change will no longer evict entries unless the org is larger than the cache size, no? So an org smaller than the cache size that deleted a team will still return stale data?

fiat-github/src/main/java/com/netflix/spinnaker/fiat/roles/github/GithubTeamsUserRolesProvider.java
rfarrjr: This is a single item cache of all members in the github org.   Will add a comment.

fiat-github/src/main/java/com/netflix/spinnaker/fiat/roles/github/GitHubProperties.java
rfarrjr: Previously loadRoles for each users would make 1 call to verify org membership, 1 call to get all teams for the org and 1 call per team to check team membership.   So for 100 users and 100 orgs the initial cache load would result in 10020 calls to github which would fire around once an hour.   

With the introduction of the teamsCache and the membersCache along with the change of the CacheKey to just be the teamId the same number of users and teams result in 102 calls to github.   As a result I felt it would be ok to reduce the CacheTTL used for the refresh eligibility.

rfarrjr:
That is true.  My understanding of refreshAfterWrite is that the entry will remain in the cache until it is requested again.  Upon the refresh eligibility the entry would be reloaded.  If that fails then it would return the stale value.   If a team was deleted from the org the teams cache would reflect this and the entry would no longer be accessed in the teamMembershipCache.  In the event it was the 404 would result in setting an empty value for that entry.    

I could add an expiresAfterWrite with an expiry 2x the membershipCacheTTLSeconds which would clear it out.

rfarrjr:
Nudge @stewchen @ttomsu .

ttomsu:
Sorry for the delay.

devorbitus:
@rfarrjr you ended up making the cache configurable but I am curious what settings you ended up using for your large number of teams

rfarrjr:
```yaml
auth:
  groupMembership:
    github:
      membershipCacheTTLSeconds: 1200
      paginationValue: 100
      membershipCacheTeamsSize: 2000
```

@devorbitus  This is our current config.

devorbitus:
This is great, thanks for sharing! I think this may help resolve https://github.com/spinnaker/spinnaker/issues/5966

PR #248, opened 2018-07-12
feat(core): Support allowing default access to all unknown applications

ajordens:
An unknown application is one that does not have explicitly specified
permissions. It may exist in `front50` or it may not.

Can be enabled by `fiat.allowAccessToUnknownApplications: true`.

### DIFF
--- a/fiat-roles/src/main/java/com/netflix/spinnaker/fiat/providers/DefaultApplicationProvider.java
+++ b/fiat-roles/src/main/java/com/netflix/spinnaker/fiat/providers/DefaultApplicationProvider.java
@@ -17,6 +17,7 @@
 package com.netflix.spinnaker.fiat.providers;
 
 import com.netflix.spinnaker.fiat.model.resources.Application;
+import com.netflix.spinnaker.fiat.model.resources.Permissions;
 import com.netflix.spinnaker.fiat.providers.internal.ClouddriverService;
 import com.netflix.spinnaker.fiat.providers.internal.Front50Service;
 import org.springframework.beans.factory.annotation.Autowired;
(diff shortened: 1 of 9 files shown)


fiat-roles/src/main/java/com/netflix/spinnaker/fiat/providers/DefaultApplicationProvider.java
```diff
@@ -58,6 +62,15 @@ public DefaultApplicationProvider(Front50Service front50Service, ClouddriverServ
           .filter(app -> !appByName.containsKey(app.getName()))
           .forEach(app -> appByName.put(app.getName(), app));
 
+      if (allowAccessToUnknownApplications) {
```
ajordens: idea is that if we're allowing access to unknown applications (aka any that don't explicitly have permissions specified), than we don't need to explicitly include them in the user permissions.

this significantly reduces the size of a `UserPermission` when you're dealing with many thousands of applications of which only a few actually have permissions.

fiat-web/src/main/java/com/netflix/spinnaker/fiat/controllers/AuthorizeController.java
```diff
@@ -83,6 +82,7 @@ public AuthorizeController(Registry registry,
         .values()
         .stream()
         .map(UserPermission::getView)
+        .map(u -> u.setAllowAccessToUnknownApplications(configProps.isAllowAccessToUnknownApplications()))
```
ajordens: The `getAll()` endpoint.

fiat-web/src/main/java/com/netflix/spinnaker/fiat/controllers/AuthorizeController.java
```diff
@@ -251,4 +229,11 @@ public void getUserAuthorization(@PathVariable String userId,
 
     return Optional.ofNullable(userPermission);
   }
+
+  private UserPermission.View getUserPermissionView(String userId) {
+    return getUserPermissionOrDefault(userId)
+        .orElseThrow(NotFoundException::new)
+        .getView()
+        .setAllowAccessToUnknownApplications(configProps.isAllowAccessToUnknownApplications());
```
ajordens: Opt'd to just tack it on to the `UserPermission.View` rather than actually store it as an attribute of the `UserPermission` itself. 

Seemed alright?

PR #249, opened 2018-07-12
feat(roles): Adding dynomite support

robzienert:
Quick implementation for dynomite. How do key changes impact things in fiat? Will things just move along ok, or do I need to come up with a key migration story (done this before).


ajordens:
Some concerns on how the key change would play with OSS --- feels like it should be migrated

robzienert:
Now also includes a bug fix for revoking roles as part of put permission operations.

I'll work on a key migrator and make it optional (but enabled by default). I figure we'd be fine just declaring bankruptcy on our fiat permissions early on, but depending on how quickly we move, it may be useful.

robzienert:
Removed the keys change, since we can just do the hashtags in the `prefix` config. Simplifies the PR quite a lot to basically a bug fix + a configuration class.

PR #253, opened 2018-07-20
feat(api): Allow `admin` to access any resource regardless of permission

ajordens:


ajordens:
FYI @ttomsu (not sure if you're back yet or not)

ttomsu:
@PerGon - I think he originally added the admin functionality

PR #254, opened 2018-07-26
feat(api): FiatPermissionEvaluator reliablity

cfieber:
Adds retry support for calls to fiat in FiatPermissionEvaluator
Invalidates cached users if they contain legacyFallback permissions

### DIFF
--- a/fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatPermissionEvaluator.java
+++ b/fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatPermissionEvaluator.java
@@ -37,10 +37,13 @@ import org.springframework.security.access.PermissionEvaluator;
 import org.springframework.security.core.Authentication;
 import org.springframework.security.core.context.SecurityContextHolder;
 import org.springframework.stereotype.Component;
+import org.springframework.util.backoff.BackOffExecution;
+import org.springframework.util.backoff.ExponentialBackOff;
 
 import java.io.Serializable;
 import java.util.Arrays;
 import java.util.Set;
+import java.util.concurrent.Callable;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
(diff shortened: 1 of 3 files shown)


fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatPermissionEvaluator.java
```diff
@@ -59,14 +62,69 @@
...
+    RetryHandler NOOP = new RetryHandler() {};
+  }
+
+  static class ExponentialBackoffRetryHandler implements RetryHandler {
+    private final long maxBackoff;
+    private final long initialBackoff;
+    private final double backoffMultiplier;
+
+    public ExponentialBackoffRetryHandler() {
+      this(10000, 500, 1.5);
```
cfieber: retry for 10 seconds, at 500, 750, 1125, 1687, 2531, 3796 ms intervals

fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatPermissionEvaluator.java
```diff
@@ -59,14 +62,69 @@
...
 
+  private final RetryHandler retryHandler;
+
+
+  interface RetryHandler {
+    default <T> T retry(String description, Callable<T> callable) throws Exception {
+      return callable.call();
+    }
+
+    RetryHandler NOOP = new RetryHandler() {};
```
cfieber: noops for the test gods

fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatPermissionEvaluator.java
```diff
@@ -176,22 +234,23 @@ public void invalidatePermission(String username) {
...
+      log.error(
+              "Cannot get whole user permission for user {}, reason: {}",
+              username,
+              exception.get().getMessage()
+      );
       id = id.withTag("legacyFallback", legacyFallback.get());
     }
 
+    if (legacyFallback.get()) {
+      permissionsCache.invalidate(username);
```
cfieber: this is new - if we returned a legacyFallback, invalidate the entry in the cache so we will try again next time we ask for the user

fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatPermissionEvaluator.java
dreynaud: > retry for 10 seconds, at 500, 750, 1125, 1687, 2531, 3796 ms intervals

would be even better as a code comment!

dreynaud: Can we make this configurable via a property?

fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatPermissionEvaluator.java
```diff
@@ -176,22 +234,23 @@ public void invalidatePermission(String username) {
...
-          exception.get().getMessage()
-      );
-    }
-
     Id id = getPermissionCounterId
         .withTag("cached", cacheHit.get())
         .withTag("success", successfulLookup.get());
 
     if (!successfulLookup.get()) {
+      log.error(
```
dreynaud: can we also log the `exception` object?

PR #255, opened 2018-08-02
fix(roles): Avoid filtering when `allowAccessToUnknownApplications` is enabled

ajordens:
This flag was originally added to allow sites with large numbers of
applications to avoid needlessly serializing the same set of applications
for each user.

Unfortunately it doesn't play nice when an application was explicitly
filtered out because a user was not granted access to it!

This PR will no longer filter applications but rather include them in
the permission with _no_ authorizations.

It only applies when `allowAccessToUnknownApplications` is enabled.

```
[
  {
    "name": "clouddriver",
    "authorizations": [
      "READ",
      "WRITE"
    ]
  },
  {
    "name": "orca",
    "authorizations": [

    ]
  }
]
```

vs

```
[
  {
    "name": "clouddriver",
    "authorizations": [
      "READ",
      "WRITE"
    ]
  }
]
```

### DIFF
--- a/fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatPermissionEvaluator.java
+++ b/fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatPermissionEvaluator.java
@@ -42,6 +42,8 @@ import org.springframework.util.backoff.ExponentialBackOff;
 
 import java.io.Serializable;
 import java.util.Arrays;
+import java.util.Collections;
+import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.Callable;
 import java.util.concurrent.ExecutionException;
@@ -289,8 +291,13 @@ public class FiatPermissionEvaluator implements PermissionEvaluator {
     Function<Set<? extends Authorizable>, Boolean> containsAuth = resources ->
         resources
             .stream()
-            .anyMatch(view -> view.getName().equalsIgnoreCase(resourceName) &&
-                view.getAuthorizations().contains(authorization));
+            .anyMatch(view -> {
+              Set<Authorization> authorizations = Optional.ofNullable(
+                  view.getAuthorizations()
+              ).orElse(Collections.emptySet());
+
+              return view.getName().equalsIgnoreCase(resourceName) && authorizations.contains(authorization);
+            });
 
     switch (resourceType) {
       case ACCOUNT:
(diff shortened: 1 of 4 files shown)


fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatPermissionEvaluator.java
```diff
@@ -289,8 +291,13 @@ private boolean permissionContains(UserPermission.View permission,
     Function<Set<? extends Authorizable>, Boolean> containsAuth = resources ->
         resources
             .stream()
-            .anyMatch(view -> view.getName().equalsIgnoreCase(resourceName) &&
-                view.getAuthorizations().contains(authorization));
+            .anyMatch(view -> {
+              Set<Authorization> authorizations = Optional.ofNullable(
```
ajordens: The null was basically a guard for test cases, in reality it'll be deserialized to an empty `Set<Authorization>`. 

I wasn't planning on proactively pushing a new `fiat-api` out to clouddriver/front50/orca.

fiat-api/src/test/groovy/com/netflix/spinnaker/fiat/shared/FiatPermissionEvaluatorSpec.groovy
```diff
@@ -198,10 +190,32 @@ class FiatPermissionEvaluatorSpec extends Specification {
   }
 
   @Unroll
-  def "should allow access to unknown applications"() {
-    given:
-    def authentication = new PreAuthenticatedAuthenticationToken("testUser", null, [])
+  def "should deny access to an application that has an empty set of authorizations"() {
```
ajordens: Just asserting that we're a-ok with resources that exist on a `UserPermission` but don't themselves have any authorizations.

fiat-roles/src/main/java/com/netflix/spinnaker/fiat/providers/DefaultApplicationProvider.java
```diff
@@ -76,4 +84,22 @@ public DefaultApplicationProvider(Front50Service front50Service,
...
+       * By default, the `BaseProvider` parent methods will filter out any applications that the authenticated user does
+       * not have access to.
+       *
+       * This is incompatible with `allowAccessToUnknownApplications` which implicitly grants access to any unknown (or
+       * filtered) applications.
+       *
+       * In this case, it is appropriate to just return all applications and allow the subsequent authorization checks
+       * to determine whether read, write or nothing should be granted.
+       */
+      return getAll();
```
ajordens: Put another way ... 

`allowAccessToUnknownApplications` is enabled, applications will be returned but with an empty set of authorizations if the user was not granted access.

`allowAccessToUnknownApplications` is disabled, only applications that the user was granted access to will be returned and have at least one authorization.

The later remains the default out-of-the-box behavior.

ajordens:
FYI @ttomsu --- just a note that we're running fiat in a mode where we _don't_ track all 6000 applications for each authenticated user ... rather we just track the ones that have permission requirements defined and allow default access to anything not listed.

Defaults haven't changed.

PR #256, opened 2018-08-14
feat(api): Surface context around what resource was denied access

ajordens:
In response to user feedback that seeing just a `403` error lacked
sufficient context to determine exactly what permission was lacking,
particularly when a pipeline or orchestration failed.

The chosen response messaging mimics what clouddriver is already
doing.

https://github.com/spinnaker/clouddriver/blob/master/clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/deploy/DescriptionValidator.groovy#L29

```
{
  "error": "Forbidden",
  "message": "Access denied to application clouddriver",
  "status": 403,
  "timestamp": 1534222128232
}
```

### DIFF
--- a/fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatAccessDeniedExceptionHandler.java
+++ b/fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatAccessDeniedExceptionHandler.java
@@ -1,0 +1,40 @@
+/*
+ * Copyright 2018 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.fiat.shared;
+
+import org.springframework.http.HttpStatus;
+import org.springframework.security.access.AccessDeniedException;
+import org.springframework.web.bind.annotation.ControllerAdvice;
+import org.springframework.web.bind.annotation.ExceptionHandler;
+
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import java.io.IOException;
+
+@ControllerAdvice
+public class FiatAccessDeniedExceptionHandler {
+  @ExceptionHandler(AccessDeniedException.class)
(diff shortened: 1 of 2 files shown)


fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatPermissionEvaluator.java
```diff
@@ -175,20 +176,13 @@ public boolean hasPermission(Authentication authentication,
     }
 
     UserPermission.View permission = getPermission(getUsername(authentication));
-    return permissionContains(permission, resourceName.toString(), r, a);
-  }
+    boolean hasPermission = permissionContains(permission, resourceName.toString(), r, a);
 
-  private String getUsername(Authentication authentication) {
```
ajordens: just moved this method down below with all the other `private` functions.

fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatPermissionEvaluator.java
```diff
@@ -175,20 +176,13 @@ public boolean hasPermission(Authentication authentication,
...
-      Object principal = authentication.getPrincipal();
-      if (principal instanceof User) {
-        username = ((User) principal).getUsername();
-      } else if (StringUtils.isNotEmpty(principal.toString())) {
-        username = principal.toString();
-      }
-    }
-    return username;
+    authenticationFailure.set(
+        hasPermission ? null : new AuthenticationFailure(a, r, resourceName.toString())
```
ajordens: thread local should be cleared on each pass --- success or failure

ajordens:
Thoughts @ttomsu / @cfieber? 

We've gotten some fairly pointed feedback that pipelines failing with just a `403` error message is confusing.

Mimic'd the error message someone would get when running a clouddriver operation against an account or application they lacked access to.

Any fear that leaking an application / account name is problematic?

jrsquared:
😍

fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatAccessDeniedExceptionHandler.java
```diff
@@ -0,0 +1,40 @@
...
+import javax.servlet.http.HttpServletResponse;
+import java.io.IOException;
+
+@ControllerAdvice
+public class FiatAccessDeniedExceptionHandler {
+  @ExceptionHandler(AccessDeniedException.class)
+  public void handleAccessDeniedException(AccessDeniedException e,
+                                          HttpServletResponse response,
+                                          HttpServletRequest request) throws IOException {
+    String errorMessage = FiatPermissionEvaluator.getAuthenticationFailure().map(a ->
```
ttomsu: I think this should be `AuthorizationFailure`, not authentication - especially since the message you're outputting includes "Access denied..."

ajordens:
Yar. You are totally correct ... muscle memory often fails me on
authentication vs authorization.

PR #257, opened 2018-08-29
fix(ldap): Return a new ExternalUser from multiLoadRoles in LDAP

dibyom:
For the LDAP provider, we were overwriting external roles if they were already set. This resulted in fiat service accounts getting synced with no roles at all.

Huge thanks to @emptywee for helping debug this!

@lwander we should cherry pick this into 1.9

Fixes https://github.com/spinnaker/spinnaker/issues/3214


lwander:
Yes please cherry-pick

PR #263, opened 2018-09-28
feat(roles): Configurable `__unrestricted_user__` roles

ajordens:
Delegates loading to the `UserRolesProvider` with the default remaining
an empty set.

We have a custom `UserRolesProvider` at Netflix and would like the
ability to specify one or more additional roles for the anonymous user.


ajordens:
@ttomsu The tricky bit with our environment is that we actually have a certificate that identifies as `anonymous` ... and it's roles are getting removed whenever the sync happens. We've got some WRITE-restricted applications that have READ open to the `anonymous` role.

I've run this locally with our `fiat-nflx` and it's doing what I expect it do. 

No change in behavior at all for OSS.

PR #288, opened 2018-11-07
fix(sync): Stop syncing service accounts in roleProvider

dibyom:
This was causing the `roles/sync` endpoint to timeout when there were a large number of service accounts. 
e.g.: https://github.com/spinnaker/spinnaker/issues/3562

### DIFF
--- a/fiat-roles/src/main/java/com/netflix/spinnaker/fiat/roles/UserRolesSyncer.java
+++ b/fiat-roles/src/main/java/com/netflix/spinnaker/fiat/roles/UserRolesSyncer.java
@@ -17,6 +17,7 @@
 package com.netflix.spinnaker.fiat.roles;
 
 import com.diffplug.common.base.Functions;
+import com.google.common.collect.Maps;
 import com.netflix.appinfo.InstanceInfo;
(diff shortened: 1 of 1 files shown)


fiat-roles/src/main/java/com/netflix/spinnaker/fiat/roles/UserRolesSyncer.java
```diff
@@ -211,13 +203,32 @@ public long updateUserPermissions(Map<String, UserPermission> permissionsById) {
...
+  }
+
+  public long updateUserPermissions(Map<String, UserPermission> permissionsById) {
+    if (permissionsById.remove(UnrestrictedResourceConfig.UNRESTRICTED_USERNAME) != null) {
+      permissionsRepository.put(permissionsResolver.resolveUnrestrictedUser());
+      log.info("Synced anonymous user role.");
+    }
+
+    Map<String, UserPermission> serviceAccounts = getServiceAccountsAsMap();
+   // Service account permissions are already fetched from Front50
```
ttomsu: nit: formatting

PR #291, opened 2018-11-13
Revert "fix(sync): Stop syncing service accounts in roleProvider"

ajordens:
Reverts spinnaker/fiat#288


ajordens:
FYI @dibyom --- revert of the previous PR

ttomsu:
What was the problem?

ajordens:
We have numerous service accounts with role membership that in-turn grants
access to a restricted account.

With calls to .resolve() removed, these service accounts were no longer
granted the account resources.

PR #296, opened 2018-12-03
feat(serviceAccounts/orMode): Change group behaviour of service accounts

gardleopard:
The current logic of service accounts is that if the service account is member of group a and b, the user that wants to use the service account needs to be a member of group a AND b. This patch changes this AND to an OR.

The rationale behind this is that when we have applications that are managed by multiple teams, they should be able to use a service account in the trigger and edit the pipelines without requiring the one to edit the pipelines to be a member of all the teams involved.

To enable this behaviour one sets `fiat.role.orMode=true` in the fiat
configuration

This is an opt in version of https://github.com/spinnaker/fiat/pull/229


ttomsu:
LGTM - would like to see @dibyom 's take on it, since he's done a lot of work with service accounts.

PR #299, opened 2018-12-07
feat(application/add): add endpoint for updating applications

gardleopard:
This endpoint enables front50 to tell fiat about application permission updates.
Without this, new applications might not be available to the user after
creation.

Depends on https://github.com/spinnaker/front50/pull/403

### DIFF
--- a/fiat-roles/src/main/java/com/netflix/spinnaker/fiat/providers/BaseProvider.java
+++ b/fiat-roles/src/main/java/com/netflix/spinnaker/fiat/providers/BaseProvider.java
@@ -36,9 +36,9 @@ import java.util.stream.Collectors;
 @Slf4j
 public abstract class BaseProvider<R extends Resource> implements ResourceProvider<R> {
 
-  private static final Integer CACHE_KEY = 0;
+  protected static final Integer CACHE_KEY = 0;
 
-  private Cache<Integer, Set<R>> cache = buildCache(20);
+  protected Cache<Integer, Set<R>> cache = buildCache(20);
 
   @Override
   @SuppressWarnings("unchecked")
@@ -74,6 +74,11 @@ public abstract class BaseProvider<R extends Resource> implements ResourceProvid
       }
       throw new ProviderException(this.getClass(), e.getCause());
     }
+  }
+
+  @Override
+  public Set<R> addItem(R item) {
+    return getAll();
   }
 
   @Autowired
--- a/fiat-roles/src/main/java/com/netflix/spinnaker/fiat/providers/DefaultApplicationProvider.java
+++ b/fiat-roles/src/main/java/com/netflix/spinnaker/fiat/providers/DefaultApplicationProvider.java
@@ -92,7 +92,7 @@ public class DefaultApplicationProvider extends BaseProvider<Application> implem
     Set<Application> applicationSet = getAll();
     if (applicationSet == null) {
       log.error("The application cache is empty. (This is ok if you do not have any applications at all.)");
-      applicationSet = new HashSet<>(); // new HashSet if this is the first application that gets added to spinnaker
+      applicationSet = new HashSet<>();
     }
     Set<Application> applicationSetMutable = new HashSet<>(applicationSet);
     if (applicationSetMutable.contains(application)) {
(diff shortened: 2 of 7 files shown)


fiat-roles/src/main/java/com/netflix/spinnaker/fiat/providers/BaseProvider.java
```diff
@@ -76,6 +76,11 @@
     }
   }
 
+  @Override
+  public Set<R> addItem(R item) {
```
ttomsu: Is this method required to compile? If not, I'd suggest removing, since it doesn't do what the method name suggests.

fiat-roles/src/main/java/com/netflix/spinnaker/fiat/providers/DefaultApplicationProvider.java
```diff
@@ -85,6 +87,24 @@ public DefaultApplicationProvider(Front50Service front50Service,
     }
   }
 
+  @Override
+  public Set<Application> addItem(Application application) {
+    Set<Application> applicationSet = getAll();
+    if (applicationSet == null) {
+      log.error("The application cache is empty. (This is ok if you do not have any applications at all.)");
```
ttomsu: If this could be ok, I'd make it `log.warn`

fiat-roles/src/main/java/com/netflix/spinnaker/fiat/providers/DefaultApplicationProvider.java
```diff
@@ -85,6 +87,24 @@ public DefaultApplicationProvider(Front50Service front50Service,
...
+      applicationSet = new HashSet<>(); // new HashSet if this is the first application that gets added to spinnaker
```
ttomsu: comment isn't needed

fiat-roles/src/main/java/com/netflix/spinnaker/fiat/providers/internal/Front50Service.java
```diff
@@ -77,6 +78,19 @@ public void afterPropertiesSet() throws Exception {
         }).execute();
   }
 
+  public Application getApplicationPermissions(String appName) {
+    try {
+      return front50Api.getApplicationPermissions(appName);
+    } catch (RetrofitError retrofitError) {
+      log.warn("Could not retrieve application {}, status code was {}",
```
ttomsu: might want to include `retrofitError.getResponse().getReason()`

fiat-web/src/main/java/com/netflix/spinnaker/fiat/controllers/ApplicationsController.java
```diff
@@ -0,0 +1,65 @@
...
+
+  private final Front50Service front50Service;
+
+  public ApplicationsController(ResourceProvider<Application> applicationProvider, Front50Service front50Service) {
+    this.applicationProvider = applicationProvider;
+    this.front50Service = front50Service;
+  }
+
+  @RequestMapping(value = "/{applicationName:.+}", method = RequestMethod.PUT)
+  public ResponseEntity<Map<String, String>> updateApplication(@PathVariable String applicationName) {
```
ttomsu: I think you can do this without the Spring-specific `ResponseEntity`

Something like:

```
public Map updateApplication(@PathVariable String applicationName, HttpServletResponse response) {
  if (app == null) {
    response.setStatus(HttpServletResponse.SC_CONFLICT)
    return Collections.singletonMap("status", "failure")
  }
  ...
  return Collections.singletonMap("status", "success")
}
```

fiat-roles/src/main/java/com/netflix/spinnaker/fiat/providers/BaseProvider.java
gardleopard: It has to be there to compile.

fiat-roles/src/main/java/com/netflix/spinnaker/fiat/providers/DefaultApplicationProvider.java
gardleopard: For all cases where you have an app or more in spinnaker this is a case you should not reach. So I think it is an error. This should only appear when you do not have any apps at all and you create the first one and the cache is younger than 20 seconds.

gardleopard:
Its probably better to do it like here https://github.com/spinnaker/fiat/pull/303

PR #303, opened 2018-12-17
feat(application/add): add endpoint for updating applications and service accounts

gardleopard:
This endpoint enables front50 to tell fiat about permission updates.
Without this, new permissions might not be available to the user after
creation.

This is a more generic way of solving https://github.com/spinnaker/fiat/pull/299 and adds service account management as well


ttomsu:
Going to close this since it's pretty old.

PR #324, opened 2019-01-30
chore(core): Clean up dependencies

ezimanyi:


fiat-roles/fiat-roles.gradle
```diff
@@ -23,15 +23,11 @@ dependencies {
 
   compile spinnaker.dependency("bootActuator")
   compile spinnaker.dependency("bootWeb")
+  compile spinnaker.dependency("googleApiClient")
+  compile spinnaker.dependency("korkDynomite")
+  compile spinnaker.dependency("korkDynomite")
```
dibyom: Duplicate?

ezimanyi: Yup, good catch

ezimanyi:
@ttomsu : Each commit has a pretty detailed description, so this is what the squashed commit message will look like.
```
* chore(google): Clean fiat-google-groups dependencies

Use the spinnaker-dependencies provided google-api-client. Also
add in bootWeb and commonsLang that are directly used but were
only included transitively.  Finally, remove a number of compileOnly
dependencies that are not used.

* chore(github): Clean fiat-github dependencies

Add some required dependencies that were only included transitively,
and remove a few dependencies that were not used.

* chore(roes): Clean fiat-roles dependencies

Push google-api-client to spinnaker-dependencies, and remove
two unused dependencies.

* chore(roles): Remove durian dependency

Was looking into whether to bump the durian dependency, but it's
only used in a single place and the same functionality exists
directly in Java 8, so just removed the dependency.

* fix(core): Remove duplicate dependency
```

ttomsu:
Two things:

1. I know I don't go looking into the commit messages, so even if they are there, the PR description is always my starting point, and I'm sure the starting point of many others.

2. IIRC, when squashing commits, only the first commit gets included in the changelog. You'd be better off doing a "rebase and merge", where each commit goes in individually, and should show up in the changelog as such. My understanding of the changelog generator may be wrong there, so I'd be glad to be proved wrong.

ezimanyi:
Fair point on the extra click from PR to commit message.  For the changelog, it does actually dig into the description of the squished commit and includes the individual descriptions.

PR #335, opened 2019-02-21
fix(redis): Fix leaked embedded redis from tests

ezimanyi:
Embedded redis is currently defined as a prototype bean in the test config; this means that a new embedded redis will be generated any time a redis is need in the auotwiring config. We're currently only destroying one of them (the one directly asked for in the test class). Instead, make this a singleton bean with a spring-managed destruction method, and use the same bean everywhere.


ezimanyi:
I found the [change](https://github.com/spinnaker/fiat/commit/ce0704d4dae23e197560245c6e00952776a2bba3#diff-0f0f7a36d531aca70bb65e107ef035a5) that made this a prototype.  It looks like that change was fixing an issue where tests were closing each other's redis connections---I think that rather than make the bean a prototype, just removing the destruction from the tests themselves and having Spring manage the bean is a better fix.

PR #343, opened 2019-03-07
chore(boot2): add order annotation, cannot have duplicate order now

emjburns:
I'm the process of converting echo to boot 2. Turns out we can't have multiple beans with the same order. Currently `WebSecurityConfigurerAdapter` has an order of 100. Adding an order annotation with a number around there so that things don't conflict


emjburns:
Never mind. Doing this in kork on `ActuatorEndpointsConfiguration`

PR #355, opened 2019-03-13
feat(buildservices): Permission support for build services (CI's)

jervi:
See https://github.com/spinnaker/igor/pull/374 for the description


jervi:
Added igor enabled flag

PR #356, opened 2019-03-14
chore(syncroles): Improve partial sync of roles

xavileon:
Doing a full sync of all roles takes a linear amount of time with respect to the number
of users. This patch only updates the users that are affected by a role change. If no roles
are specified, then a full sync is done as we were doing before.


xavileon:
@dibyom @jervi @amanya

fiat-roles/src/main/java/com/netflix/spinnaker/fiat/roles/UserRolesSyncer.java
```diff
@@ -31,6 +31,8 @@
 import com.netflix.spinnaker.fiat.providers.ResourceProvider;
 import com.netflix.spinnaker.kork.eureka.RemoteStatusChangedEvent;
 import com.netflix.spinnaker.kork.lock.LockManager;
+
+import java.util.*;
```
dibyom: Minor nit: Can we undo this wildcard import and add back the java.util imports as it were before?

xavileon: sure, done!

PR #361, opened 2019-03-26
Boot2

dibyom:


fiat-web/src/main/java/com/netflix/spinnaker/fiat/config/ResourcesConfig.java
```diff
@@ -62,6 +65,10 @@
   @Setter
   private String clouddriverEndpoint;
 
+  @Value("${services.igor.baseUrl}")
```
robzienert: This needs to be `${services.igor.base-url}` now.

dibyom: Done!

PR #363, opened 2019-04-01
Remove potentially expensive Redis health indicator

robfletcher:


marchello2000:
lgtm, we could make it `conditional(false)` for now but for expediency this makes sense

robfletcher:
Right, we want to revisit but when things aren't on fire.

PR #366, opened 2019-04-10
feat(admin): Implement isAdmin check in FiatPermissionEvaluator.

dibyom:
The previous implementation is used in some Front50 batch APIs so we can't get rid of it just yet!

### DIFF
--- a/fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatPermissionEvaluator.java
+++ b/fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatPermissionEvaluator.java
@@ -350,6 +350,15 @@ public class FiatPermissionEvaluator implements PermissionEvaluator {
     return true; // TODO(ttomsu): Chosen by fair dice roll. Guaranteed to be random.
   }
 
+
+  public boolean isAdmin(Authentication authentication) {
+    if (!fiatStatus.isEnabled()) {
+      return true;
+    }
+
+    return getPermission(getUsername(authentication)).isAdmin();
+  }
+
   public class AuthorizationFailure {
     private final Authorization authorization;
     private final ResourceType resourceType;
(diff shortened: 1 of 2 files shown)


fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatPermissionEvaluator.java
```diff
@@ -350,6 +350,15 @@ public boolean isAdmin() {
     return true; // TODO(ttomsu): Chosen by fair dice roll. Guaranteed to be random.
   }
 
+
+  public boolean isAdmin(Authentication authentication) {
+    if (!fiatStatus.isEnabled()) {
+      return true;
+    }
+
+    return getPermission(getUsername(authentication)).isAdmin();
```
ajordens: I think it's possible that `getPermission` could return null (or raise an exception) ... worth a null check and possibly a try/catch that would return false rather than propagate the exception.

fiat-api/src/test/groovy/com/netflix/spinnaker/fiat/shared/FiatPermissionEvaluatorSpec.groovy
```diff
@@ -251,6 +251,26 @@ class FiatPermissionEvaluatorSpec extends Specification {
...
 
+  def "should support isAdmin check for a user"() {
+    when:
+    1 * fiatService.getUserPermission("testUser") >> {
+      return new UserPermission.View()
+          .setApplications(Collections.emptySet())
+          .setAdmin(true)
+    }
+    then:
+    evaluator.isAdmin(authentication)
```
ajordens: believe you've got your when and then backward ... at least how I would normally write a test like this.

Alternatively:

```
given:
1 * fiatService.getUserPermission("testUser") >> {
      return new UserPermission.View()
          .setApplications(Collections.emptySet())
          .setAdmin(isAdmin)
    }

expect:
evaluator.isAdmin(authentication) == expectedIsAdmin

where:
isAdmin || expectedIsAdmin
false   || false
true    || true
```

ajordens:
The `front50` batch api usage is a little bogus ... it's calling `isAdmin` but that's always returning `true` so not really enforcing anything.

ajordens:
Thanks for the fast follow-up on getting this change in.

fiat-api/src/test/groovy/com/netflix/spinnaker/fiat/shared/FiatPermissionEvaluatorSpec.groovy
dibyom: :+1:

PR #373, opened 2019-04-15
feat(roles): Add Execute as a type of Authorization

dibyom:
Only adds the new type. Follow up PRs will:
- Migrate existing applications by adding execute permissions to users who can currently execute pipelines (PR:https://github.com/spinnaker/front50/pull/516)
- Switch to using Execute for checks during a pipeline trigger.

Part of https://github.com/spinnaker/spinnaker/issues/3554

### DIFF
--- a/fiat-core/src/main/java/com/netflix/spinnaker/fiat/model/resources/Permissions.java
+++ b/fiat-core/src/main/java/com/netflix/spinnaker/fiat/model/resources/Permissions.java
@@ -106,8 +106,7 @@ public class Permissions {
                .collect(Collectors.toSet());
   }
 
-  @VisibleForTesting
-  protected List<String> get(Authorization a) {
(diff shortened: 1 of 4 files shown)


fiat-core/src/main/java/com/netflix/spinnaker/fiat/model/resources/Permissions.java
```diff
@@ -106,8 +106,7 @@ public boolean isEmpty() {
                .collect(Collectors.toSet());
   }
 
-  @VisibleForTesting
-  protected List<String> get(Authorization a) {
+  public List<String> get(Authorization a) {
```
dibyom: This makes writing the migration easier as there isn't a easy way now to get a list of all groups who have READ or WRITE access.

PR #375, opened 2019-04-16
feat(api): Allow access to unknown applications when legacyFallback=true

ajordens:
Recent testing of our legacy fallback support did not actually work
correctly when `fiat` was unavailable.

The original intention of `legacyFallback: true` was to support a
fallback to the authorization model that existed pre-fiat.

This model only supported WRITE restricted accounts and expected that
the authorization context was propagated between services with the
`X-SPINNAKER-*` headers.

This PR is an attempt to do a better job for manually triggered pipelines
(which should have some accounts specified) as well as read operations
performed via the UI.

### DIFF
--- a/fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatPermissionEvaluator.java
+++ b/fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatPermissionEvaluator.java
@@ -232,7 +232,7 @@ public class FiatPermissionEvaluator implements PermissionEvaluator {
                             .map(a -> new Account().setName(a))
                             .collect(Collectors.toSet())
                     )
-            ).setLegacyFallback(true);
+            ).setLegacyFallback(true).setAllowAccessToUnknownApplications(true);
           }
         }).call();
       });
@@ -253,10 +253,6 @@ public class FiatPermissionEvaluator implements PermissionEvaluator {
               exception
       );
       id = id.withTag("legacyFallback", legacyFallback.get());
-    }
-
-    if (legacyFallback.get()) {
-      permissionsCache.invalidate(username);
     }
 
     registry.counter(id).increment();


fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatPermissionEvaluator.java
```diff
@@ -255,10 +255,6 @@ public void invalidatePermission(String username) {
       id = id.withTag("legacyFallback", legacyFallback.get());
     }
 
-    if (legacyFallback.get()) {
```
ajordens: By default, the `permissionsCache` is already evicting after 20s.

When `fiat` is unavailable, we run into a situation where `clouddriver` is calling `hasPermission()` in a tight loop with each iteration taking 10s of retries. 

Removing `permissionsCache.invalidate()` results in only one set of retries.

ajordens: This primarily improves the situation for READs (and WRITES from manually triggered pipelines) when `fiat` is unavailable.

Will follow-up with something that'll be smarter about handling cron triggered/anonymous pipelines.

cfieber: what about also not putting the generated legacy permissions into the cache? do the legacy fallback building behaviour outside the cacheloader on exception?

cfieber: that might have the same effect as invalidating though

cfieber: trying to recall why I stuck the invalidate in there in the first place - it was a dark time. There is probably a case in the fallback building that we should fail on looking closer

cfieber: this little gem in the fallback builder seems pretty suspect: `AuthenticatedRequest.getSpinnakerUser().orElse("anonymous")` what should we be doing with the permission if there is no X-SPINNAKER-USER?

ajordens: It's perhaps masking situations where we're not correctly propagating `X-SPINNAKER-USER` already, seems like something else to investigate -- (cron'd pipelines aren't explicitly initiated with `anonymous`)

ajordens: I do have some ideas around allowing source of truth systems to better define their fallbacks:
- `clouddriver` owns accounts and should be able to override X-SPINNAKER-ACCOUNTS 
- `igor` owns jenkins and should be able to define what masters are available based off local config (there's no current fallback for this)

ajordens: We've got a number of spots where we're using `AuthenticatedRequest.getSpinnakerUser().orElse("anonymous")`, it's probably a sign that we need better visibility around areas where we're not propagating the `X-SPINNAKER-*` headers _or_ not even setting an auth context when starting an execution (ie. cron triggers).

I think there are opportunities (or needs) to have better fallback calculations than just `X-SPINNAKER-ACCOUNTS`. Complicated by the fact that permissions are now being checked in most of the services regardless of whether they own the data or not.

PR #377, opened 2019-04-17
fix(dependencies): dependency cleanup

cfieber:
* removes lots of cargo-culted dependency declarations
* trims dependencies in fiat-core and fiat-api to minimize the
  impact consuming projects
* guava caches -> caffeine


fiat-api/src/test/groovy/com/netflix/spinnaker/fiat/shared/FiatPermissionEvaluatorSpec.groovy
```diff
@@ -103,46 +103,18 @@ class FiatPermissionEvaluatorSpec extends Specification {
                                                .setAuthorizations([Authorization.READ] as Set)] as Set)
     upv.setServiceAccounts([new ServiceAccount.View().setName(svcAcct)
                                                      .setMemberOf(["foo"])] as Set)
+    fiatService.getUserPermission("testUser") >> upv
```
cfieber: changes in here are related to the change to Caffeine cache not being deterministic about the number of items (we have it set to maxItems = 0 for tests, but eviction back down to 0 happens async). The places where it was asserting the number of calls against the mock were brittle but also were not relevant to the tests - they were effectively stub calls anyway so the tests got a bit simpler as a result

PR #378, opened 2019-04-17
Revert "fix(front50): Update Front50 cache periodically and serve live calls from cache. (#351)"

dibyom:
This reverts commit 4ea2d71a717b2366acae34170823596ccf43dcc4.

Having this cache along with the cache in `BaseProvider` was causing a lot of user role syncs to use stale data from Front50. This should alleviate the problems in 1.13 e.g. https://github.com/spinnaker/spinnaker/issues/4276

Ideally, I think we should try to refresh the `BaseProvider` cache with live data before a role sync starts but I'll make that a follow up PR.

### DIFF
--- a/fiat-roles/src/main/java/com/netflix/spinnaker/fiat/providers/internal/Front50Service.java
+++ b/fiat-roles/src/main/java/com/netflix/spinnaker/fiat/providers/internal/Front50Service.java
@@ -25,6 +25,7 @@ import lombok.Getter;
 import lombok.extern.slf4j.Slf4j;
 import org.springframework.beans.factory.InitializingBean;
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.scheduling.annotation.Scheduled;
 
 import java.util.List;
 import java.util.concurrent.atomic.AtomicReference;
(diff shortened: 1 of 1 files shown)


fiat-roles/src/main/java/com/netflix/spinnaker/fiat/providers/internal/Front50Service.java
```diff
@@ -50,40 +51,53 @@ public Front50Service(Front50Api front50Api) {
...
   }
 
   public List<Application> getAllApplicationPermissions() {
-    return applicationCache.get();
+    return new SimpleJava8HystrixCommand<>(
+        GROUP_KEY,
+        "getAllApplicationPermissions",
+        () -> {
+          applicationCache.set(front50Api.getAllApplicationPermissions());
+          healthTracker.success();
```
ajordens: if this call fails when fiat first starts, it'll never go healthy?

dibyom: It will eventually since the `UserRoleSyncer` is scheduled to run every 10 minutes. This was actually the behavior until very recently. I was going to add some retries to the initial cache prime in a follow up but can add it as part of this as well!

ajordens: 👍 

(I see `refreshCache()` back in here but nothing apparently calling it -- aside from that, this looks good to me)

We've ended up lowering the sync interval here from 10min to 2min (and actually feature flagged off the explicit sync'ing from `front50` to `fiat` due to it being an incident causing denial of service). 

We'll likely have to revisit that later bit but aren't yet creating applications with permissions initially set.

dibyom: Its scheduled to run every `"${fiat.front50RefreshMs:30000}`. But I also added to the initial call in `afterPropertiesSet()`

dibyom:
@spinnakerbot cherry-pick 1.13

PR #379, opened 2019-04-17
feat(api): Support for dynamic configuration of the Fiat retry policies

ajordens:


fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatClientConfigurationProperties.java
```diff
@@ -47,8 +56,38 @@
 
   @Data
   class RetryConfiguration {
+    private DynamicConfigService dynamicConfigService;
```
robzienert: If we go down this path, do we just want to use the fastprops propetysource instead?

cfieber: I think the issue is that `@ConfigurationProperties` don't change values to reflect the underlying value in the Environment, they get bound once on startup, so the pattern of injecting DynamicConfigService into the configuration properties bean gives us a yaml-bound-but-current-to-the-environment ConfigurationProperties bean but without maybe the super-magic that comes along with hooking up to spring-cloud-config

PR #382, opened 2019-04-19
feat(api): Support overriding the connect/read timeouts for `FiatService`

ajordens:
Currently these timeouts are inherited from normal service <-> service
communication.

The preference would be that an individual call to `fiat` would timeout
before the overarching service rpc times out.

Shorter timeouts would allow the existing retry behavior in
`FiatPermissionEvaluator` to be more effective.


asher:
LGTM

PR #385, opened 2019-04-22
fix(api): Support `legacyFallback` for `BUILD_SERVICE`

ajordens:
When `service.fiat.legacyFallback` is `true`, this will implicitly allow
access to all build services in the event that `fiat` is unavailable.


ajordens:
@dibyom We are currently going through an exercise to understand the situation if (and when) `fiat` were to become unavailable.

Since our adoption, we've always been supporting a `legacyFallback` model wherein we effectively fall back to the state that existed prior to fiat (account restrictions only).

This may need revisiting at some point _but_ it's a bare minimum that we'd like to see supported  for now.

ajordens:
Also thinking that there may be a model where source of truth systems (ie. jenkins, clouddriver) could supply their own fallbacks (or an extension point allowing a user to supply their own fallback logic).

ajordens:
@dibyom If enabled, it would. 

We've bounced around the idea for an optional extension point (likely in igor) that could potentially blacklist certain build services even for fallbacks.

PR #387, opened 2019-04-24
WIP: feat(roles): Add fallback behavior for EXECUTE authz checks

dibyom:
For backwards compatibility, fallback to the existing behavior i.e.
READ users have access to EXECUTE. The `services.fiat.executeFallbackToWrite`
flag allows changing this behavior to allow WRITE users instead.

Part of https://github.com/spinnaker/spinnaker/issues/3554


ajordens:
This is no longer relevant given spinnaker/fiat#390 right?

dibyom:
Yeah. Closing in favor of #390

PR #388, opened 2019-04-24
upgrades fiat to boot2 and new BOM dependency management

cfieber:
gradle 5.4, spring-boot 2 dependencies, and kork-bom as the source of version opinions


build.gradle
```diff
@@ -74,20 +56,34 @@ allprojects {
       jvmArgs '-agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=7103'
     }
   }
+
+  //workaround nebula oss default 1.7 opinion..
+  plugins.withType(JavaBasePlugin) {
```
cfieber: this can be removed w/ spinnaker-gradle-project 6.0.0

fiat-api/fiat-api.gradle
```diff
@@ -15,33 +15,22 @@
...
-  compileOnly spinnaker.dependency("kork")
-  compileOnly spinnaker.dependency("lombok")
-  annotationProcessor spinnaker.dependency("lombok")
-  compileOnly spinnaker.dependency("okHttp")
-  compileOnly spinnaker.dependency("okHttpUrlconnection")
-  compileOnly spinnaker.dependency("okHttpApache")
-  compileOnly spinnaker.dependency("retrofit")
-  compileOnly spinnaker.dependency("retrofitJackson")
-  compileOnly spinnaker.dependency("spectatorApi")
+  implementation "org.springframework:spring-aop"
```
cfieber: these all are actually runtime dependencies. With BOM based dependency opinions we won't run into problems on version alignment so it is safe to make these implementation rather than compileOnly

fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatAuthenticationConfig.java
```diff
@@ -98,40 +96,16 @@ FiatWebSecurityConfigurerAdapter fiatSecurityConfig(FiatStatus fiatStatus) {
     private final FiatStatus fiatStatus;
 
     private FiatWebSecurityConfigurerAdapter(FiatStatus fiatStatus) {
+      super(true);
```
cfieber: this disables defaults, our auth filter chain for fiat is mostly about having the FiatAuthenticationFilter in there and falling back to anonymous

fiat-core/src/main/java/com/netflix/spinnaker/fiat/model/resources/Permissions.java
```diff
@@ -152,15 +141,16 @@ public Builder add(Authorization a, List<String> groups) {
     }
 
     public Permissions build() {
-      ImmutableMap.Builder<Authorization, List<String>> builder = ImmutableMap.builder();
+      final Map<Authorization, List<String>> perms = new HashMap<>();
```
cfieber: unguava's this

fiat-core/src/main/java/com/netflix/spinnaker/fiat/model/resources/ResourceType.java
```diff
@@ -33,11 +32,9 @@
   }
 
   // TODO(ttomsu): This is Redis-specific, so it probably shouldn't go here.
-  public static ResourceType parse(@NonNull String pluralOrKey) {
-    if (pluralOrKey.contains(":")) {
-      pluralOrKey = StringUtils.substringAfterLast(pluralOrKey, ":");
-    }
-    String singular = StringUtils.removeEnd(pluralOrKey, "s");
+  public static ResourceType parse(@Nonnull String pluralOrKey) {
+    pluralOrKey = pluralOrKey.substring(pluralOrKey.lastIndexOf(':') + 1);
```
cfieber: un commons-langs this

fiat-core/src/main/java/com/netflix/spinnaker/fiat/model/resources/Role.java
```diff
@@ -48,9 +46,9 @@ public Role(String name) {
     this.setName(name);
   }
 
-  public Role setName(String name) {
-    if (StringUtils.isEmpty(name)) {
-      throw new IllegalArgumentException("name cannot be null or empty");
+  public Role setName(@Nonnull String name) {
+    if (name.isEmpty()) {
+      throw new IllegalArgumentException("name cannot be empty");
```
cfieber: lombok will catch the null case via `@Nonnull` annotation

gradle.properties
```diff
@@ -1,2 +1,3 @@
-org.gradle.parallel=true
-includeProviders=file,github,google-groups,ldap
\ No newline at end of file
+org.gradle.parallel=false
```
dibyom: This should probably be true?

PR #389, opened 2019-04-25
feat(google): Add support for using group email as a Fiat role name

perek:
The G-Suite resolver attempts to use the Google Group Name to set the role, this is undesirable for a number of reasons. This PR adds support for specifying which key to use for role mappings. This change allows for backwards compatibility as the role mapper will use any number of RoleSource's to be specified, allowing for a period of migration from Name -> Email. @dibyom


Dockerfile
```diff
@@ -6,7 +6,7 @@ COPY . workdir/
 
 WORKDIR workdir
 
-RUN GRADLE_USER_HOME=cache ./gradlew buildDeb -x test && \
+RUN GRADLE_USER_HOME=cache ./gradlew buildDeb -I gradle/init-publish.gradle -x test && \
```
perek: I had to fix this in order to get the docker to build

fiat-google-groups/src/test/groovy/com/netflix/spinnaker/fiat/roles/google/GoogleDirectoryUserRolesProviderSpec.groovy
```diff
@@ -0,0 +1,72 @@
...
+import com.google.api.services.admin.directory.model.Groups;
+import spock.lang.Specification
+
+class GoogleDirectoryUserRolesProviderSpec extends Specification {
+    GoogleDirectoryUserRolesProvider.Config config = new GoogleDirectoryUserRolesProvider.Config()
+
+    def "should read google groups"() {
+        setup:
+        config.domain = "test.com"
+        def group = new Group()
```
dibyom: Can get rid of the setters: ` def group = new Group(name: "test", email: "test@test.com")`

perek: i am not great with groovy :P

perek:
@dibyom I have that commit on my branch, i was getting build errors for missing task 'buildDeb' when running docker build, unless i need to cherry pick another commit in.

dibyom:
Yeah `buildDeb` would have needed the init-publish script. But with that init script was deleted in https://github.com/spinnaker/fiat/pull/388/files#diff-31322cd550ad389e6ddcef0d11863332 and needs to be replaced with `-PenabledPublishing=true` instead. 

I'd be ok with taking the Dockerfile changes out of this PR and making it a separate commit.

PR #397, opened 2019-04-29
fix(api): Rebuild the legacy fallback if it has no account grants

ajordens:
This PR preserves the debouncing capability of the existing
permissionsCache but attempts to provide a fresher fallback in the
event that a "no accounts" fallback was previously cached.

The "no accounts" fallback will result if the X-SPINNAKER-* headers
are not propagated for a given api request (or that api request
originated from an scheduled system operation and thus was not
authenticated to begin with).


ajordens:
Approved per spinnaker/fiat#396

PR #400, opened 2019-05-01
feat(api): Log any access granted as part of a legacy fallback.

ajordens:


fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatPermissionEvaluator.java
```diff
@@ -177,6 +177,16 @@ public boolean hasPermission(
     authorizationFailure.set(
         hasPermission ? null : new AuthorizationFailure(a, r, resourceName.toString()));
 
+    if (permission != null && permission.isLegacyFallback() && hasPermission) {
+      // log any access that was granted as part of a legacy fallback.
+      if (a == Authorization.READ) {
```
robzienert: One-letter var names are sadface.

ajordens: indeed they are

ajordens: thanks travis

fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatPermissionEvaluator.java
```diff
@@ -177,6 +177,16 @@ public boolean hasPermission(
...
+        // purposely logging at 'debug' as 'READ' will be sufficiently more verbose
```
dreynaud: > sufficiently

did you mean significantly?

ajordens: tbh I did mean sufficiently but I speak Canadian English so maybe I'm in the wrong.

PR #405, opened 2019-05-08
fix(ldap): Disable LDAP health indicator

ezimanyi:
Fixes spinnaker/spinnaker#4346.  When we discussed a couple of weeks ago, I didn't realize there was a way to turn of the health indicator from the code (rather than from config).  Given it's possible to do that, it's probably easier to do this and then revert it once we pull in the upstream fix.

There is currently an NPE in the LDAP health indicator that has been fixed but not yet released which is causing fiat to never appear healthy. Until we can pull in the upstream fix, disable the LDAP health indicator.


ezimanyi:
This should only disable the health check (which is currently broken anyway), so the effect would be that fiat would still show healthy even if the downstream LDAP is down.

PR #409, opened 2019-05-21
fix(apps): Clear resource provider caches to ensure up-to-date data

ncknt:
When syncing roles, Fiat recomputes unrestricted roles for all the apps. Apps are cached for 20s (by default). This causes issues when creating an app (which will also do a sync) within 20s of another sync: the new app will not be reflected in the unrestricted role and result in 403s.

This fix clears the cache of the various providers to ensure they're refetched from Front50/Clouddriver.

This should address https://github.com/spinnaker/spinnaker/issues/4421

### DIFF
--- a/fiat-roles/src/main/java/com/netflix/spinnaker/fiat/permissions/DefaultPermissionsResolver.java
+++ b/fiat-roles/src/main/java/com/netflix/spinnaker/fiat/permissions/DefaultPermissionsResolver.java
@@ -91,6 +91,13 @@ public class DefaultPermissionsResolver implements PermissionsResolver {
     return getUserPermission(user.getId(), combo);
   }
 
+  @Override
+  public void clearCache() {
+    for (ResourceProvider provider : resourceProviders) {
+      provider.clearCache();
+    }
+  }
+
   private boolean resolveAdminRole(Set<Role> roles) {
     List<String> adminRoles = fiatAdminConfig.getAdmin().getRoles();
     return roles.stream().map(Role::getName).anyMatch(adminRoles::contains);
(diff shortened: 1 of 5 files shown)


ncknt:
@ezimanyi @ajordens Could you take a look? Took the easy way to fix, could also pass a `refreshCache` boolean to the `getAll` method but that will be a much bigger PR.

fiat-roles/src/main/java/com/netflix/spinnaker/fiat/permissions/DefaultPermissionsResolver.java
```diff
@@ -64,6 +64,10 @@
 
   @Override
   public UserPermission resolveUnrestrictedUser() {
+    // Ensure we're going to reload app and service account definitions
+    for (ResourceProvider provider : resourceProviders) {
```
dibyom: Can we invalidate the cache before we start a user role sync instead of in the middle of it ? i.e. somewhere here:
https://github.com/spinnaker/fiat/blob/master/fiat-roles/src/main/java/com/netflix/spinnaker/fiat/roles/UserRolesSyncer.java#L145

ezimanyi:
This looks reasonable to me once @dibyom 's comment is resolved.

fiat-roles/src/main/java/com/netflix/spinnaker/fiat/permissions/DefaultPermissionsResolver.java
ncknt: Sure. Moved to `UserRoleSyncer`.

ncknt:
@dibyom I've updated the PR. Let me know if that works for you.

PR #412, opened 2019-05-22
 fix(fiat): Fiat Doesn't Support Structured Logging #4409

neelimamanubolu:
Fiat Doesn't Support Structured Logging #4409:

I have added logstashEncoder in fiat-core.gradle and made the log file initialised in lcass path.Able to check the logs.

### DIFF
--- a/fiat-core/fiat-core.gradle
+++ b/fiat-core/fiat-core.gradle
@@ -3,7 +3,7 @@ dependencies {
   implementation "com.fasterxml.jackson.core:jackson-annotations"
   implementation "com.google.code.findbugs:jsr305"
   implementation "org.slf4j:slf4j-api"
-
+  implementation "net.logstash.logback:logstash-logback-encoder"
   testImplementation "org.springframework.boot:spring-boot"
   testImplementation "com.fasterxml.jackson.core:jackson-databind"
 }


fiat-core/fiat-core.gradle
```diff
@@ -1,5 +1,6 @@
 dependencies {
 
+  implementation "net.logstash.logback:logstash-logback-encoder:4.11"
```
ezimanyi: I think this should leave the version unspecified and use the version from `kork` rather than specifying a specific version.

neelimamanubolu: Hi Eric, as the issue itself explians that doesnt support structured logging, fiat cant take kork dependency like all other service. Hence we specified the logger with version.

ezimanyi: I understand that you need to pull in the `logstash-logback-encoder` to fix this issue, but my question is why you are explicitly specifying the version in `fiat-core.gradle`.  In general (as you'll see for other dependencies in that file) we leave the version unspecified and rely on the versions specified in [spinnaker-dependencies](https://github.com/spinnaker/kork/blob/master/spinnaker-dependencies/spinnaker-dependencies.gradle) in `kork` to actually set the version.

neelimamanubolu: Ok.. will update the fiat-core.gradle without version and submit new PR for this. Thanks

neelimamanubolu:
Need to do some changes in previous PR

PR #419, opened 2019-06-06
fix(redis): mitigate JedisPool depletion

cfieber:
Fixes redis interactions in RedisPermissionRepository that can result in
depleting the redis pool.

Specifically:
* converts smembers calls to sscan calls, and gets/returns an object
  from the pool to do so.
* removes nested calls to redisClientDelegate that were holding on
  to pool objects
* partitions lookup of users into smaller batches
* partitions role lookup

Additionally, uses InstrumentedJedisPool to enable metrics collection
on redis interactions.

### DIFF
--- a/fiat-roles/src/main/java/com/netflix/spinnaker/fiat/config/RedisConfig.java
+++ b/fiat-roles/src/main/java/com/netflix/spinnaker/fiat/config/RedisConfig.java
@@ -1,7 +1,9 @@
 package com.netflix.spinnaker.fiat.config;
 
+import com.netflix.spectator.api.Registry;
 import com.netflix.spinnaker.kork.jedis.JedisClientDelegate;
 import com.netflix.spinnaker.kork.jedis.RedisClientDelegate;
+import com.netflix.spinnaker.kork.jedis.telemetry.InstrumentedJedisPool;
 import java.net.URI;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.commons.lang3.StringUtils;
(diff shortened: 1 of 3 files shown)


fiat-roles/src/main/java/com/netflix/spinnaker/fiat/config/RedisConfig.java
```diff
@@ -62,6 +65,9 @@ private static JedisPool createPool(
       redisPoolConfig = new GenericObjectPoolConfig();
     }
 
-    return new JedisPool(redisPoolConfig, host, port, timeout, password, database, null);
+    return new InstrumentedJedisPool(
+        registry,
+        new JedisPool(redisPoolConfig, host, port, timeout, password, database, null),
+        "fiat");
```
robzienert: I made this generic (I should deprecate `InstrumentedJedisPool`): https://github.com/spinnaker/kork/blob/master/kork-telemetry/src/main/java/com/netflix/spinnaker/kork/telemetry/InstrumentedProxy.java

Usage: https://github.com/spinnaker/kork/blob/master/kork-telemetry/src/test/groovy/com/netflix/spinnaker/kork/telemetry/InstrumentedProxySpec.groovy

fiat-roles/src/main/java/com/netflix/spinnaker/fiat/permissions/RedisPermissionsRepository.java
```diff
@@ -308,34 +299,47 @@ private UserPermission getUserPermission(String userId, RawUserPermission raw) {
...
+            p.sync();
           });
     } catch (Exception e) {
       log.error("Storage exception reading " + id + " entry.", e);
     }
   }
 
+  private Set<String> scanSet(String key) {
+    final Set<String> results = new HashSet<>();
+    final AtomicReference<String> cursor = new AtomicReference<>(ScanParams.SCAN_POINTER_START);
```
robzienert: 💯

fiat-roles/src/main/java/com/netflix/spinnaker/fiat/config/RedisConfig.java
cfieber: pending spinnaker/kork#313

cfieber: actually scratch that, InstrumentedProxy doesn't work here (it would just instrument the pool methods not the pooled object itself)

PR #424, opened 2019-06-13
fix(fiat): Fiat Doesn't Support Structured Logging #4409

neelimamanubolu:
Hi, As discussed, updtaed the required changes without version.

Please find the below link for reference ::

https://github.com/spinnaker/spinnaker/issues/4409


neelimamanubolu:
> This dependency should be in fiat-web.gradle

As refernced in spinnaker/clouddriver#3666, we should add the dependencies in fiat-core.gradle. Can u pls check once and confirm.

neelimamanubolu:
As  requested done the changes in fiat-web.gradle. Can u pls merge the PR

PR #433, opened 2019-07-01
fix(core): add back isEnabled check to hasPermission method to preven…

maggieneterval:
…t NPEs

- Prior to [this PR](https://github.com/spinnaker/fiat/pull/358), `fiatStatus.isEnabled` was called prior to trying to read anything off the (possibly null) auth object. Do this check immediately in order to prevent NPEs.


maggieneterval:
@spinnakerbot cherry-pick 1.14

PR #435, opened 2019-07-03
fix(core): add null check to auth in FiatPermissionEvaluator.getUsername

maggieneterval:


maggieneterval:
@spinnakerbot cherry-pick 1.14

PR #460, opened 2019-08-26
feat(authorization): Add resource group permissions and implement them for applications

AbdulRahmanAlHamali:
This enables us to set authorizations on the level of multiple applications, as well as set permissions on who can create applications, by adding a "*" entry with `CREATE` set only to specific groups


AbdulRahmanAlHamali:
I refactored this to make it more generic. I am using an interface called `ResourceGroup`, which could be a prefix-based group, or something else if desired. 
I also introduced an interface called `GroupResolutionStrategy`, which allows us to decide what we want to do with groups that match the resource. The default implementation is "additive", which means that we add all the groups together.

ajordens:
I’m slightly torn and not sure how I feel about the bigger picture of grouped permissions and how they would be exposed to a user and/or managed. Specifically, what would an application owner see when they look at their config?

I suspect if we were to attempt something like this at Netflix, we would have an external source of truth that could(?) be pushed into Spinnaker as needed (or pulled from front50). 

I do agree that there’s a gap around restricting who can create an application, that seems solvable. Similarly there’s also some tightening to be done around requiring an application actually exist before you can create resources/deploy/etc. 

Thoughts @ttomsu?

AbdulRahmanAlHamali:
Hello @ajordens , thank you for taking the time to reply.

Regarding the big picture, I believe grouped resources are introducing the concept of a big picture here, because they allow admins of Spinnaker to set policies that span all applications, or that divide applications into groups, while the current approach only allows us to manage each application's permissions on its own. 

The current implementation imposes problems on use-cases where one team (let's call it the admin team) is responsible for providing and managing a single Spinnaker instance that is shared among multiple teams. This model is necessary when working with heavy security compliance restrictions (this is exactly our use case). And so in this scenario, the application owner, which is the admin team, has to manage applications one-by-one, which makes it really harder to implement policies and change management. For example, let's say I have 150 applications, and a new group of users (let's call them auditor group) should be added as Read-Only to all of them to do the audits. It makes much more sense to add them to group `*` instead of adding them to each application individually. 
 
Thus, the problem is not only about creating applications, it is about full configuration management under heavier security considerations.

In addition, this implementation does not impact the per-application implementation. So if a use-case wishes to not use resource groups at all, they will not be impacted whatsoever.

cfieber:
Not totally sure how I feel about explicitly carving out `CREATE` and `DELETE` permissions vs just using the `WRITE` permission to handle the case of "I can update this thing, I can update it into existence, and I can update it out of existence" (otherwise I feel like we've set ourselves up for confusion with those authorizations and operations on cloud resources that do CREATE/DELETE but assert the WRITE permission)

I do think generally this idea makes a lot of sense though.

I just opened #472 which has some overlap with this (`GroupResolutionStrategy` could probably be a `ResourceInterceptor` from that PR)

How sold are you on differentiating CREATE and DELETE vs just using WRITE, I think that is my main hangup with this is around extending the Authorizations space.

AbdulRahmanAlHamali:
Hello @cfieber, I am not very hung on the `DELETE`, but I think the `CREATE` is definitely needed.

The problem is that the authz implementation here has an object called `UserPermissions`, which lists the applications that the user has access to, and the user's authorizations for each of those applications. With `CREATE`, we are dealing with an application that does not yet exist, and so it does not exist in the `UserPermissions` object, which makes it necessary to treat `CREATE` requests differently.

We can possibly change this whole model. I don't really understand why it's implemented that way honestly, and this can solve this problem and make the implementation much cleaner

cfieber:
@AbdulRahmanAlHamali thinking about it more I agree that CREATE does make sense for this use case, and I could potentially see us extending it onto service accounts in the future.

I'd be ok to go ahead with this pretty much as is, if we get rid of DELETE and just do a WRITE check on the application at deletion time.

We'll need to think about how these changes land in a way where we don't have a circular dependency at deploy time. I think that means that 
1. Front50 gets updated with the resourceGroups endpoint, but not asserting CREATE
2. This change lands in fiat
3. Front50 gets updated to assert CREATE

Ideally 3 could be done with a feature flag - I'll to a quick check to see if that is the case

AbdulRahmanAlHamali:
@cfieber I'm actually going to remove the resourceGroups endpoints from front50, and store the resource groups in fiat config instead. 
This will mean that each time you want to add or modify resource groups you will have to restart the fiat service, but it will also mean that you can configure them through halyard. Unfortunately the only meaningful way of configuring Spinnaker now is through halyard, so I will have to go with this.

fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatPermissionEvaluator.java
```diff
@@ -206,6 +212,36 @@ public void invalidatePermission(String username) {
...
+      Response response =
+          AuthenticatedRequest.propagate(
+                  () -> {
+                    return retryHandler.retry(
+                        String.format(
+                            "hasAuthorization for user %s on resource %s of type %s with authorization CREATE",
+                            username, resourceName, resourceType),
+                        () ->
+                            fiatService.hasAuthorization(
+                                username, resourceType, resourceName, "CREATE"));
```
cfieber: use Authorization.CREATE.name() here instead of the string

fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatPermissionEvaluator.java
```diff
@@ -206,6 +212,36 @@ public void invalidatePermission(String username) {
     permissionsCache.invalidate(username);
   }
 
+  private boolean canCreate(String username, String resourceType, String resourceName) {
+    boolean canCreate = false;
+
+    try {
+      Response response =
```
cfieber: we may want to think about including the ResourceGroup objects into the UserPermission response so that we can evaluate this without the remote call.

Given that `CREATE` is not likely on a high rps call path I'm okay with doing that later if it proves to be a problem though

fiat-core/src/main/java/com/netflix/spinnaker/fiat/model/resources/groups/ResourceGroup.java
```diff
@@ -0,0 +1,38 @@
...
+import com.fasterxml.jackson.annotation.JsonTypeInfo;
+import com.netflix.spinnaker.fiat.model.resources.Permissions;
+import com.netflix.spinnaker.fiat.model.resources.Resource;
+import com.netflix.spinnaker.fiat.model.resources.ResourceType;
+import lombok.Data;
+
+@Data
+@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, property = "resourceGroupType")
+@JsonSubTypes(@JsonSubTypes.Type(value = PrefixResourceGroup.class, name = "Prefix"))
+public abstract class ResourceGroup {
```
cfieber: Could you javadoc this class / its methods as its kind of key to the whole new feature

cfieber:
Thinking a bit more about this, I'm interested in supporting a creation check as an explicit method on FiatPermissionEvaluator rather than following the hasPermission signature.

I'm thinking of a ResourceGroup type that inspects the application attributes to determine the appropriate ResourceGroups (specifically we have something that validates owner email, and I'd like to use that as a way to key the resource group for an application, rather than relying on a naming convention for application names)

This would mean something like `canCreate(ResourceType resourceType, String resourceName, Object resource)` which could call fiat asking for resourceGroups matching type/name/resource attributes, and fiat could have an ApplicationOwnerResourceGroup.

This would likely align much better with how our existing suite of applications are named (fairly arbitrarily) and owned (owner email typically is the team's google group).

@AbdulRahmanAlHamali thoughts or concerns around this? I'm going to start experimenting a bit in that direction. (Also not proposing you make that change here, I'm willing to do that)

cfieber:
> Thinking a bit more about this, I'm interested in supporting a creation check as an explicit method on FiatPermissionEvaluator rather than following the hasPermission signature.
> 
> I'm thinking of a ResourceGroup type that inspects the application attributes to determine the appropriate ResourceGroups (specifically we have something that validates owner email, and I'd like to use that as a way to key the resource group for an application, rather than relying on a naming convention for application names)
> 
> This would mean something like `canCreate(ResourceType resourceType, String resourceName, Object resource)` which could call fiat asking for resourceGroups matching type/name/resource attributes, and fiat could have an ApplicationOwnerResourceGroup.
> 
> This would likely align much better with how our existing suite of applications are named (fairly arbitrarily) and owned (owner email typically is the team's google group).
> 
> @AbdulRahmanAlHamali thoughts or concerns around this? I'm going to start experimenting a bit in that direction. (Also not proposing you make that change here, I'm willing to do that)

immediate potential concern to is that we can't build/assert resource groups without having the resource contents available, however as we crawl authorized objects when building up UserPermission I do think we have those available..

cfieber:
@AbdulRahmanAlHamali https://github.com/coveord/fiat/pull/1 addresses the couple minor comments I had

fiat-roles/src/main/java/com/netflix/spinnaker/fiat/providers/internal/resourcegroups/BaseGroupResolutionStrategy.java
```diff
@@ -0,0 +1,67 @@
...
+ */
+
+package com.netflix.spinnaker.fiat.providers.internal.resourcegroups;
+
+import com.netflix.spinnaker.fiat.model.resources.Permissions;
+import com.netflix.spinnaker.fiat.model.resources.Resource;
+import com.netflix.spinnaker.fiat.model.resources.groups.ResourceGroup;
+import java.util.Set;
+import java.util.stream.Collectors;
+
```
AbdulRahmanAlHamali: @cfieber I am not very happy with what I'm doing here.
I don't like that we give the resolution strategy the whole, unfiltered groups, and it takes care of the filtering. I would rather give it pre-filtered groups, and just get rid of that `BaseResolutionStrategy` class, but I don't know where to put the filtering logic. Do you have any suggestions?

cfieber: I was thinking a bit about this, and with the move away from front50 storing the ResourceGroups I don't think we need to keep a single jackson-deserializable hierarchy anymore.

I think we could make a ResourceGroupProvider (similar to ResourceProvider), make those Spring components, and they could independently get configuration (via `@ConfigurationProperties`) for the rules they need.

This is maybe similar to where I was headed in https://github.com/spinnaker/fiat/pull/472 except s/ResourceInterceptor/ResourceGroupProvider/ and also supporting the lookup of permissions for a resource that does not yet exist.

We can then autowire in a collection of all ResourceGroupProviders that exist, and each of them can augment permissions for whatever resources they support. This also has the added benefit of being able to extend/add new ResourceGroupProviders via plugins (right now the ResourceGroupType / jackson deserialization makes that kind of impossible).

AbdulRahmanAlHamali: Hmmmm I think I've understood most of what you said except the last paragraph.
Having a `ResourceGroupProvider` is indeed a good idea. It is already kind of implemented right now in `FiatResourceGroupConfig`, but I can move the logic out to a separate provider and make things more generic. Let me see what I can do

PR #463, opened 2019-09-03
fix(logging): log at higher level for some auth failures

marchello2000:
When a param is missing from the `hasPermission` call, log it louder


fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatPermissionEvaluator.java
```diff
@@ -151,8 +151,8 @@ public boolean hasPermission(
       return true;
     }
     if (resourceName == null || resourceType == null || authorization == null) {
-      log.debug(
-          "Permission denied due to null argument. resourceName={}, resourceType={}, "
+      log.warn(
```
marchello2000: there is prob benefit of unifying with a similar approach as https://github.com/spinnaker/fiat/pull/463/files#diff-5a4302861533ea9742ab03634d0d1a78R177 but then we would also have to fix up the `@controlleradvice` exception handler in front50..

PR #472, opened 2019-09-19
feat(roles): Introduce ResourceInterceptor extension point.

cfieber:
A resource interceptor allows for modification of the resources that come back from ResourceProviders.

### DIFF
--- a/fiat-core/src/main/java/com/netflix/spinnaker/fiat/model/resources/BaseAccessControlled.java
+++ b/fiat-core/src/main/java/com/netflix/spinnaker/fiat/model/resources/BaseAccessControlled.java
@@ -24,8 +24,6 @@ import lombok.extern.slf4j.Slf4j;
 abstract class BaseAccessControlled<R extends BaseAccessControlled>
     implements Resource.AccessControlled {
 
-  abstract R setPermissions(Permissions p);
-
   /**
    * Legacy holdover where setting `requiredGroupMembership` implied both read and write
    * permissions.
--- a/fiat-roles/src/main/java/com/netflix/spinnaker/fiat/providers/BaseProvider.java
+++ b/fiat-roles/src/main/java/com/netflix/spinnaker/fiat/providers/BaseProvider.java
@@ -23,6 +23,9 @@ import com.google.common.util.concurrent.UncheckedExecutionException;
 import com.netflix.spinnaker.fiat.config.ProviderCacheConfig;
 import com.netflix.spinnaker.fiat.model.resources.Resource;
 import com.netflix.spinnaker.fiat.model.resources.Role;
+import java.lang.reflect.ParameterizedType;
+import java.util.Collections;
+import java.util.List;
 import java.util.Set;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
@@ -37,6 +40,35 @@ public abstract class BaseProvider<R extends Resource> implements ResourceProvid
   private static final Integer CACHE_KEY = 0;
 
   private Cache<Integer, Set<R>> cache = buildCache(20);
+
+  private final List<ResourceInterceptor> resourceInterceptors;
+
+  public BaseProvider() {
+    this(Collections.emptyList());
+  }
+
+  public BaseProvider(List<ResourceInterceptor> resourceInterceptors) {
+    Class<R> resourceType = resolveResourceType(getClass());
+    this.resourceInterceptors =
+        resourceInterceptors == null
+            ? Collections.emptyList()
+            : resourceInterceptors.stream()
+                .filter(ri -> ri.supports(resourceType))
+                .collect(Collectors.toList());
+  }
+
+  // VisibleForTesting
+  List<ResourceInterceptor> inspectResourceInterceptors() {
+    return Collections.unmodifiableList(resourceInterceptors);
+  }
+
+  @SuppressWarnings("unchecked")
+  private static <R extends Resource> Class<R> resolveResourceType(
+      Class<? extends BaseProvider> clazz) {
+    ParameterizedType type = (ParameterizedType) clazz.getGenericSuperclass();
+    Class<R> resourceType = (Class<R>) type.getActualTypeArguments()[0];
+    return resourceType;
+  }
 
   @Override
   @SuppressWarnings("unchecked")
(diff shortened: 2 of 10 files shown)


fiat-core/src/main/java/com/netflix/spinnaker/fiat/model/resources/BaseAccessControlled.java
```diff
@@ -24,8 +24,6 @@
 abstract class BaseAccessControlled<R extends BaseAccessControlled>
     implements Resource.AccessControlled {
 
-  abstract R setPermissions(Permissions p);
```
cfieber: moved this to the AccessControlled interface (all the concrete subclasses already explicitly support it)

cfieber:
concrete use case we are experimenting with is configuring some groups that by default get read-only access to all resources. For example we want SREs and central platform type teams to be able to see any application, regardless of the configured permissions on the application

fiat-roles/src/test/groovy/com/netflix/spinnaker/fiat/providers/BaseProviderSpec.groovy
```diff
@@ -111,12 +114,68 @@ class BaseProviderSpec extends Specification {
...
+
+    then:
+    resources.size() == 1
+    resources[0].name == "reqGroup1and2"
+    resources[0].permissions.getAuthorizations(["group1", "group2"]) == [Authorization.READ] as Set
+    !resources[0].permissions.get(Authorization.WRITE)
+    !resources[0].permissions.get(Authorization.EXECUTE)
+  }
+
+  class ReadOnlyResourceInterceptor implements ResourceInterceptor {
```
cfieber: stupid ResourceInterceptor example - mutate all resources to read-only

ajordens:
❤️ 

I _really_ like this.

fiat-roles/src/main/java/com/netflix/spinnaker/fiat/providers/ResourceInterceptor.java
```diff
@@ -0,0 +1,30 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.fiat.providers;
+
+import com.netflix.spinnaker.fiat.model.resources.Resource;
+import java.util.Set;
+
+public interface ResourceInterceptor {
```
robzienert: Would like to see interfaces & their methods get documentation.

cfieber:
going to close this and rework on top of #460 once that lands

fiat-roles/src/main/java/com/netflix/spinnaker/fiat/providers/BaseProvider.java
```diff
@@ -90,5 +122,13 @@ public void clearCache() {
     cache.invalidate(CACHE_KEY);
   }
 
+  protected Set<R> populateCache() throws ProviderException {
```
ttomsu: Probably want to make this private unless there's a good reason subclasses need to meddle with the interceptors.

fiat-roles/src/main/java/com/netflix/spinnaker/fiat/providers/BaseProvider.java
```diff
@@ -38,6 +41,35 @@
...
+  private final List<ResourceInterceptor> resourceInterceptors;
+
+  public BaseProvider() {
+    this(Collections.emptyList());
+  }
+
+  public BaseProvider(List<ResourceInterceptor> resourceInterceptors) {
+    Class<R> resourceType = resolveResourceType(getClass());
+    this.resourceInterceptors =
+        resourceInterceptors == null
```
plumpy: IMO, accepting `null` (especially in component constructors) just seems like encouraging bad behavior. People shouldn't be handing you a null object, and I don't think there's any reason that would happen as the code is written currently, so just don't allow it.

plumpy:
I added my own proposal (which is very similar, but different in what I think are imporant ways) in #478. Curious to get any feedback...

fiat-roles/src/main/java/com/netflix/spinnaker/fiat/providers/BaseProvider.java
cfieber: Fair enough - first pass through this I tried to keep everything non breaking (hence the extra constructors etc) as I was working on it, but I can definitely unwind that

fiat-roles/src/main/java/com/netflix/spinnaker/fiat/providers/ResourceInterceptor.java
```diff
@@ -0,0 +1,30 @@
...
+  default boolean supports(Class<? extends Resource> type) {
```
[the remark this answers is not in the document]
cfieber: The NoopResourceInterceptor is taking advantage, but I could definitely just move these defaults there.

cfieber:
>  how do you decide which ResourceInterceptors come first

If an interceptor is opinionated on which order it should run in, it can implement Ordered or use `@Order`

> Since the new code you want to add is only really relevant for applications

My internal use case applies to all `Resource.AccessControlled` hence the reflection shenanigoats and ability to support multiple resource types.

> permissions stored on the resource in front50 always get loaded first, but why should that be the case

At this point I'd say mostly because it keeps existing Spinnaker installations consistent and allows optionally layering on this new functionality as a way of applying rules-based configurations after the fact. I can't think of a use-case currently where I'd want to throw out the existing permission model so I guess I was opting for maybe less sweeping of a change 

> I added my own proposal (which is very similar, but different in what I think are imporant ways) in #478.

I will take another look there, I think there is some good stuff but I don't think it currently gives a great story on how to extend the current model (I could see how to do it, but it feels more invasive that the interceptor approach to me)

plumpy:
> I can't think of a use-case currently where I'd want to throw out the existing permission model

Perhaps users want to store their permissions in LDAP or some other IAM system, alongside the roles.

Why does the existing permission model get special treatment? Just because it was written first? That doesn't seem like a great reason to me.

If this new permissions model wants to build on top of the existing model, that's still easy... just have your `NamePrefixApplicationPermissionProvider` delegate to the `Front50ApplicationPermissionProvider` to get the initial list of permissions and then build on top of that. But now you're explicitly defining that relationship in code instead of just letting it happen.

> it feels more invasive that the interceptor approach to me

The interceptor approach feels _so_ invasive to me. It allows anyone to define an interceptor that does basically anything to the list of resources... add some, remove some, modify them in any way... there's essentially no constraints on it whatsoever. I think that level of flexibility is both unnecessary and likely to be abused...

cfieber:
> The interceptor approach feels so invasive to me. It allows anyone to define an interceptor that does basically anything to the list of resources... 

True enough. Let me poke around a bit to see if there is some middle ground. As an example, I'd maybe like to be able to use both the name prefix config, on top of front50 config, also with an internal extension. I want to make sure there is a way to potentially assemble all those, and I worry if the extension mechanism becomes at the ApplicationPermissionProvider level that I might not have the ability to choose whether for example NamePrefixApplicationProvider decided it needed to delegate to Front50 or not

Noodling! Thanks for all the discussion on this

PR #474, opened 2019-09-25
fix(error): Improve access denied error messaging

jonsie:
This is the first part of the change to improve front50's access denied error messaging.  After this change we will replace front50's `AccessDeniedExceptionHandler` with `FiatAccessDeniedExceptionHandler`.

Here are examples of the different error message contexts:

```
//Non service account example:
{
  "timestamp": "2019-09-25T18:41:21.833+0000",
  "status": 403,
  "error": "Forbidden",
  "message": "Access denied to application smalleytest - required authorization: WRITE"
}

//Service account example:
{
  "timestamp": "2019-09-25T18:44:27.137+0000",
  "status": 403,
  "error": "Forbidden",
  "message": "Access denied to serviceaccount someServiceAccount"
}
```

### DIFF
--- a/fiat-api/src/test/groovy/com/netflix/spinnaker/fiat/shared/FiatAccessDeniedExceptionHandlerSpec.groovy
+++ b/fiat-api/src/test/groovy/com/netflix/spinnaker/fiat/shared/FiatAccessDeniedExceptionHandlerSpec.groovy
@@ -1,0 +1,63 @@
+/*
+ * Copyright 2019 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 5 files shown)


fiat-api/src/test/groovy/com/netflix/spinnaker/fiat/shared/FiatAccessDeniedExceptionHandlerSpec.groovy
```diff
@@ -0,0 +1,63 @@
...
+
+    FiatPermissionEvaluator evaluator = new FiatPermissionEvaluator(
+            registry,
+            fiatService,
+            buildConfigurationProperties(),
+            fiatStatus,
+            FiatPermissionEvaluator.RetryHandler.NOOP
+    )
+
+    def "when access denied exception is handled, the descriptive error message is sent in response"() {
```
robzienert: nit: `s/the descriptive/a descriptive/`

fiat-api/src/test/groovy/com/netflix/spinnaker/fiat/shared/FiatSharedSpecification.groovy
```diff
@@ -0,0 +1,37 @@
...
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.fiat.shared
+
+import com.netflix.spectator.api.NoopRegistry
+import com.netflix.spectator.api.Registry
+import spock.lang.Specification
+
+class FiatSharedSpecification extends Specification {
```
robzienert: Seems like this should be `abstract`?

PR #478, opened 2019-10-03
refactor(fiat-roles): add a ResourcePermissionProvider interface

plumpy:
refactor(fiat-roles): add a `ResourcePermissionProvider` interface

There are two of these that are required to be provided for Fiat to
startup correctly: one for Accounts and one for Applications. There is
a default `@Configuration` that maintains the status quo by providing
default implementations that simply read the permissions directly off of
the objects, but alternate configurations can be provided that retrieve
permissions in other ways.

Curious what you think about this. The tests are almost certainly broken and more tests need to be added, etc., but it should work.

I prefer this because it allows you to define any number of policies, and those policies can define strict rules about where the permissions come from without worrying about interference from other unrelated policies. The interface is also more strict, just a mapping from `Resource` to `Permissions`, as opposed to `ResourceInterceptor` which lets implementations do any modifications they want to the `Resources`.

### DIFF
--- a/fiat-core/src/main/java/com/netflix/spinnaker/fiat/model/resources/Permissions.java
+++ b/fiat-core/src/main/java/com/netflix/spinnaker/fiat/model/resources/Permissions.java
@@ -27,8 +27,8 @@ import lombok.val;
 
 /**
  * Representation of authorization configuration for a resource. This object is immutable, which
- * makes it challenging when working with Jackson's {{ObjectMapper}} and Spring's
- * {{\@ConfigurationProperties}}. The {@link Builder} is a helper class for the latter use case.
+ * makes it challenging when working with Jackson's {@code ObjectMapper} and Spring's
+ * {@code @ConfigurationProperties}. The {@link Builder} is a helper class for the latter use case.
  */
 @ToString
 @EqualsAndHashCode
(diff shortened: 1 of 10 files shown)


AbdulRahmanAlHamali:
I really like that, very clean and flexible. @cfieber what do you think?

We still need to add something to account for creation. The `ResourcePermissionProvider` could have another method called, `getPermissionsForNewResource`, which accepts the resource as a parameter, and decided what permissions it should have (we could as well just add an `isNew` parameter to the `getPermissions` method). What do you think?

AbdulRahmanAlHamali:
Also, if I want to implement my resource group thing as a plugin, I will still be blocked because you guys have not yet provided a standardized plugin model (there is one used in `orca` but I understand it is just PoC?). So what can I do? I can port the plugin model from orca into here if you give the green light for that.

plumpy:
I think it's okay to just implement it directly in this repository until we figure out what the plugins will look like. I mostly want to make sure that whatever approach we use, we structure this in a way that it _could_ be pulled out as a plugin once that stabilizes. (It could be quite a while before that happens, so I don't want to hold this up any more than I already have.)

AbdulRahmanAlHamali:
Alright, perfect. I'll see what @cfieber has in mind, and will implement either on top of your PR or his

PR #479, opened 2019-10-03
fix(roles): Allow anonymous calls from Fiat to other Spinnaker modules

jervi:
Will suppress all the `Request GET:http://<service>/<endpoint> is missing [X-SPINNAKER-USER, X-SPINNAKER-ACCOUNTS] authentication headers and will be treated as anonymous` messages that is currently spamming the Fiat log.


23doors:
anything holding this back? would love to see it merged, these WARNs are kind of confusing

PR #480, opened 2019-10-10
do not merge: Role providers as plugins

cfieber:
do not mock me spinnakerbot

I know I am the Jerk witch lowlife, and the commit message is the trashcan that is my life.


cfieber:
airline-driven-development quick POC that identifies optional modules within a project and packages them as a zip containing the library and the additional dependencies it requires

fiat-plugins/fiat-plugins.gradle
```diff
@@ -0,0 +1,25 @@
+subprojects {
+  apply plugin: 'distribution'
+
+  distributions {
+    main {
+      contents {
+        from jar.outputs
+        from {
+          configurations.runtimeClasspath.resolvedConfiguration.resolvedArtifacts.findResults { dep ->
```
cfieber: this is the main craziness that strips out any dependency that comes along with the core app from the plugin distribution (if we wanted to go nested classloaders or anything this might make it simpler)

also worth noting - I had to rename fiat-web to fiat-main because I wanted to separate the -web code from the packaging (so I could have the packaging include all these known plugins into the distribution and figure out how to hook them up with our plugin loading story) - however we have some stuff in our gradle plugins that expects the packaging stuff to live in a project named `-web` so (don't remember why)

fiat-plugins/roles-file/roles-file.gradle
```diff
@@ -15,15 +15,10 @@
  */
 
 dependencies {
-  implementation project(":fiat-roles")
-  implementation project(":fiat-core")
-  implementation "com.fasterxml.jackson.core:jackson-core"
-  implementation "com.fasterxml.jackson.core:jackson-databind"
+  implementation project(":fiat-main")
```
cfieber: I think I could revert the dependency changes in here

I initially was looking at bringing in fiat-main as a `compileOnly` dependency, but with the dependency shenanigans I added that is no longer necessary

settings.gradle
```diff
@@ -16,16 +16,25 @@
 
 rootProject.name="fiat"
 
+def plugins = [
+  ':fiat-plugins',
+  ':fiat-plugins:kork-secrets-aws',
+  ':fiat-plugins:kork-stackdriver',
```
robzienert: I assume these bits about `kork-*` would go away once kork is plugin-ified? Not sure I like each service having kork plugins defined in their codebase.

robzienert:
I am kinda confused by this PR. I think this is solely in my lack of understanding in Gradle. How would someone build a fiat of their dreams with only the plugins they want?

robzienert:
Ping

PR #481, opened 2019-10-11
feat(aggregator): enables customization of permission aggregation as well as permission sources

cfieber:
Adds a ResourcePermissionSource - a place to get permissions from, and
ResourcePermissionProvider - the API for callers

The default ResourcePermissionProvider is an aggregator across all
ResourcePermissionSources

There is a `DefaultResourcePermissionConfiguration` that maintains the
status quo by providing default implementations that simply read the
permissions directly off of the objects, but allows opting out of some
or all of the default opinions if someone wants to customize, extend,
or override any of the default behaviour.

Additional improvements:

* refactor(fiat-roles):
    - make DefaultPermissionsResolver immutable
    - don't return null values from Permissions#get
    - clean up loadAll

closes #478 
closes #472

### DIFF
--- a/fiat-core/src/main/java/com/netflix/spinnaker/fiat/model/Authorization.java
+++ b/fiat-core/src/main/java/com/netflix/spinnaker/fiat/model/Authorization.java
@@ -18,7 +18,7 @@ package com.netflix.spinnaker.fiat.model;
 
 import java.util.Arrays;
 import java.util.Collections;
-import java.util.HashSet;
+import java.util.LinkedHashSet;
 import java.util.Set;
 
 public enum Authorization {
@@ -26,6 +26,6 @@ public enum Authorization {
   WRITE,
   EXECUTE;
 
-  public static Set<Authorization> ALL =
-      Collections.unmodifiableSet(new HashSet<>(Arrays.asList(values())));
+  public static final Set<Authorization> ALL =
+      Collections.unmodifiableSet(new LinkedHashSet<>(Arrays.asList(values())));
 }
--- a/fiat-core/src/main/java/com/netflix/spinnaker/fiat/model/Authorization.java
+++ b/fiat-core/src/main/java/com/netflix/spinnaker/fiat/model/Authorization.java
@@ -16,9 +16,8 @@
 
 package com.netflix.spinnaker.fiat.model;
 
-import java.util.Arrays;
 import java.util.Collections;
-import java.util.LinkedHashSet;
+import java.util.EnumSet;
 import java.util.Set;
 
 public enum Authorization {
@@ -27,5 +26,5 @@ public enum Authorization {
   EXECUTE;
 
   public static final Set<Authorization> ALL =
-      Collections.unmodifiableSet(new LinkedHashSet<>(Arrays.asList(values())));
+      Collections.unmodifiableSet(EnumSet.allOf(Authorization.class));
 }
--- a/fiat-core/src/main/java/com/netflix/spinnaker/fiat/model/resources/Permissions.java
+++ b/fiat-core/src/main/java/com/netflix/spinnaker/fiat/model/resources/Permissions.java
@@ -27,8 +27,8 @@ import lombok.val;
 
 /**
  * Representation of authorization configuration for a resource. This object is immutable, which
- * makes it challenging when working with Jackson's {{ObjectMapper}} and Spring's
- * {{\@ConfigurationProperties}}. The {@link Builder} is a helper class for the latter use case.
+ * makes it challenging when working with Jackson's {@code ObjectMapper} and Spring's
+ * {@code @ConfigurationProperties}. The {@link Builder} is a helper class for the latter use case.
  */
 @ToString
 @EqualsAndHashCode
@@ -90,7 +90,7 @@ public class Permissions {
   }
 
   public List<String> get(Authorization a) {
-    return permissions.get(a);
+    return permissions.getOrDefault(a, Collections.emptyList());
   }
 
   /**
--- a/fiat-core/src/main/java/com/netflix/spinnaker/fiat/model/resources/Permissions.java
+++ b/fiat-core/src/main/java/com/netflix/spinnaker/fiat/model/resources/Permissions.java
@@ -101,7 +101,7 @@ public class Permissions {
    */
   public static class Builder extends LinkedHashMap<Authorization, List<String>> {
 
-    static Permissions fromMap(Map<Authorization, List<String>> authConfig) {
+    private static Permissions fromMap(Map<Authorization, List<String>> authConfig) {
       final Map<Authorization, List<String>> perms = new EnumMap<>(Authorization.class);
       for (Authorization auth : Authorization.values()) {
         perms.put(
--- a/fiat-core/src/main/java/com/netflix/spinnaker/fiat/model/resources/Permissions.java
+++ b/fiat-core/src/main/java/com/netflix/spinnaker/fiat/model/resources/Permissions.java
@@ -34,12 +34,12 @@ import lombok.val;
 @EqualsAndHashCode
 public class Permissions {
 
-  public static Permissions EMPTY = new Permissions.Builder().build();
+  public static final Permissions EMPTY = Builder.fromMap(Collections.emptyMap());
 
   private final Map<Authorization, List<String>> permissions;
 
   private Permissions(Map<Authorization, List<String>> p) {
-    this.permissions = p;
+    this.permissions = Collections.unmodifiableMap(p);
   }
 
   /**
@@ -67,10 +67,6 @@ public class Permissions {
 
   public boolean isAuthorized(Set<Role> userRoles) {
     return !getAuthorizations(userRoles).isEmpty();
-  }
-
-  public boolean isEmpty() {
-    return permissions.isEmpty();
   }
 
   public Set<Authorization> getAuthorizations(Set<Role> userRoles) {
@@ -105,6 +101,25 @@ public class Permissions {
    */
   public static class Builder extends LinkedHashMap<Authorization, List<String>> {
 
+    static Permissions fromMap(Map<Authorization, List<String>> authConfig) {
+      final Map<Authorization, List<String>> perms = new EnumMap<>(Authorization.class);
+      for (Authorization auth : Authorization.values()) {
+        perms.put(
+            auth,
+            Optional.ofNullable(authConfig.get(auth))
+                .map(
+                    groups ->
+                        groups.stream()
+                            .map(String::trim)
+                            .filter(s -> !s.isEmpty())
+                            .map(String::toLowerCase)
+                            .collect(Collectors.toList()))
+                .map(Collections::unmodifiableList)
+                .orElse(Collections.emptyList()));
+      }
+      return new Permissions(perms);
+    }
+
     @JsonCreator
     public static Builder factory(Map<Authorization, List<String>> data) {
       return new Builder().set(data);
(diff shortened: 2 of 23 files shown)


fiat-roles/src/main/java/com/netflix/spinnaker/fiat/providers/AggregatingResourcePermissionProvider.java
```diff
@@ -0,0 +1,62 @@
...
+      if (permissions.isRestricted()) {
+        for (Authorization auth : Authorization.values()) {
+          List<String> roles = permissions.get(auth);
+          if (roles != null && !roles.isEmpty()) {
+            builder.add(auth, roles);
+          }
+        }
+      }
+    }
+
```
srekapalli: Question? Can this empty logic be with in the builder API and return permissions aggregate (Empty or otherwise)

cfieber: I will take a look, that seems reasonable to me

fiat-web/src/main/java/com/netflix/spinnaker/fiat/config/DefaultResourcePermissionConfig.java
```diff
@@ -0,0 +1,86 @@
...
+      havingValue = "default",
+      matchIfMissing = true)
+  public ResourcePermissionProvider<Application> defaultApplicationPermissionProvider(
+      List<ResourcePermissionSource<Application>> sources) {
+    return new AggregatingResourcePermissionProvider<>(sources);
+  }
+
+  @Bean
+  @ConditionalOnProperty(
+      value = "auth.permissions.source.account.resource.enabled",
```
AbdulRahmanAlHamali: should be `build-service` instead of `account`

AbdulRahmanAlHamali:
Looks Good!

A prefix implementation (or any other implementation) could very easily be built on top of that. I will branch of this and start.

My only comment is that at least the `Provider` needs another method. `getCreationPermissions`, because the decision process might be different when the operation is `CREATE`. This could also be added to the `Source`s, but not as necessarily as the provider. However, it could be done in another PR on top of that if you don't want to over-bloat this one. Let me know what you think

fiat-web/src/main/java/com/netflix/spinnaker/fiat/config/DefaultResourcePermissionConfig.java
cfieber: thanks!

fiat-web/src/main/java/com/netflix/spinnaker/fiat/config/DefaultResourcePermissionConfig.java
```diff
@@ -0,0 +1,86 @@
...
+import com.netflix.spinnaker.fiat.model.resources.Application;
+import com.netflix.spinnaker.fiat.model.resources.BuildService;
+import com.netflix.spinnaker.fiat.providers.*;
+import java.util.List;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+@Configuration
+class DefaultResourcePermissionConfig {
```
plumpy: I don't think `AggregatingResourcePermissionProvider` being the default is a good plan. I think it's going to come back to bite us in the future. This policy is essentially "every `PermissionSource` that is defined in the code base should be turned on all at the same time". It works for now (since there's only one!) but makes no sense at all when we add more and more of these.

I understand it probably makes things easier for you, but IMO if users want something besides "permissions stored in front50", they should have to manually specify what they want in a config file somewhere.

fiat-roles/src/main/java/com/netflix/spinnaker/fiat/providers/AggregatingResourcePermissionProvider.java
```diff
@@ -0,0 +1,62 @@
...
+  @Override
+  public Permissions getPermissions(T resource) {
+    Permissions.Builder builder = new Permissions.Builder();
+    for (ResourcePermissionSource<T> source : resourcePermissionSources) {
+      Permissions permissions = source.getPermissions(resource);
+      if (permissions.isRestricted()) {
+        for (Authorization auth : Authorization.values()) {
+          List<String> roles = permissions.get(auth);
+          if (roles != null && !roles.isEmpty()) {
```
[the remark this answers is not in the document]
cfieber: that wasn't actually true until I pushed 04138a8 but it definitely is now, will fix

fiat-web/src/main/java/com/netflix/spinnaker/fiat/config/DefaultResourcePermissionConfig.java
cfieber: In 92d1359 I added a DefaultResourcePermissionProvider that takes a single source only, and made that the default. Arguably that could have been AggregatingResourcePermissionProvider with a single entry list, but DefaultRPP is a simple enough class and keeps the intent more obvious.

I left aggregation in as a configuration time switch - the group based permission model and an internal integration I'm working on would both leverage this approach, so that feels like enough use-cases to warrant it being in the codebase to me.

fiat-core/src/main/java/com/netflix/spinnaker/fiat/model/Authorization.java
```diff
@@ -18,14 +18,14 @@
...
 
 public enum Authorization {
   READ,
   WRITE,
   EXECUTE;
 
-  public static Set<Authorization> ALL =
-      Collections.unmodifiableSet(new HashSet<>(Arrays.asList(values())));
+  public static final Set<Authorization> ALL =
+      Collections.unmodifiableSet(new LinkedHashSet<>(Arrays.asList(values())));
```
plumpy: `Collections.unmodifiableSet(EnumSet.allOf(Authorization.class))`

(also uses natural iteration order, presuming that's why you switched to `LinkedHashSet`)

fiat-core/src/main/java/com/netflix/spinnaker/fiat/model/resources/Permissions.java
```diff
@@ -27,19 +27,19 @@
...
- * {{\@ConfigurationProperties}}. The {@link Builder} is a helper class for the latter use case.
+ * makes it challenging when working with Jackson's {@code ObjectMapper} and Spring's
+ * {@code @ConfigurationProperties}. The {@link Builder} is a helper class for the latter use case.
  */
 @ToString
 @EqualsAndHashCode
 public class Permissions {
 
-  public static Permissions EMPTY = new Permissions.Builder().build();
+  public static final Permissions EMPTY = Builder.fromMap(Collections.emptyMap());
```
plumpy: I think `new Builder().build()` is a little more straightforward personally, but feel free to ignore...

fiat-core/src/main/java/com/netflix/spinnaker/fiat/model/resources/Permissions.java
```diff
@@ -105,6 +101,25 @@ public boolean isEmpty() {
    */
   public static class Builder extends LinkedHashMap<Authorization, List<String>> {
 
+    static Permissions fromMap(Map<Authorization, List<String>> authConfig) {
```
plumpy: make this `private`

fiat-core/src/main/java/com/netflix/spinnaker/fiat/model/resources/Permissions.java
cfieber: The reason I moved away from new Builder().build() was because build() returns this instance if it builds an empty permission object

plumpy: Oh yeah, duh, good point thanks :grin:

fiat-roles/src/main/java/com/netflix/spinnaker/fiat/providers/AccessControlledResourcePermissionSource.java
```diff
@@ -19,13 +19,16 @@
...
 
 public final class AccessControlledResourcePermissionSource<T extends Resource.AccessControlled>
     implements ResourcePermissionSource<T> {
 
   @Override
-  public Permissions getPermissions(T resource) {
-    return Optional.ofNullable(resource.getPermissions())
+  @Nonnull
+  public Permissions getPermissions(@Nonnull T resource) {
+    return Optional.ofNullable(resource)
```
cfieber: he marks resource `@Nonnull` then `Optional.ofNullable`'s it... trust levels showing through

fiat-core/src/main/java/com/netflix/spinnaker/fiat/model/resources/Permissions.java
cfieber: done

fiat-web/src/main/java/com/netflix/spinnaker/fiat/config/DefaultResourcePermissionConfig.java
plumpy: I guess what concerns me about is that now adding a new `PermissionSource` is potentially a breaking change... you're essentially constraining all future `PermissionSources` to act in a way that won't screw things up, even if they have nothing at all to do with you. And you're relying on future authors to even know that any new `PermissionSource` needs to safely interact with _all_ the others (including ones that Netflix has added, but which aren't in the OSS build).

This doesn't seem like a huge thing for Netflix to deal with, but I don't think it belongs in the OSS build.

AbdulRahmanAlHamali: But sources are not really interacting with each other. They are receiving a resource with well-defined permissions, and producing a resource with well-defined permission. The `ResourcePermissionProvider` is the one deciding the interactions.

In my group example, the aggregator works well whether I provide my whole group logic as one `Source`, or provide each of my groups as a separate `Source`. In both of those cases, the sources are completely isolated from the way the interaction is managed.

plumpy: Sorry, yeah, I like the source/provider distinction that Cam made, I think it's cool. My concern here is specifically with the `AggregatingResourcePermissionProvider`. That one aggregates from every single `ResourcePermissionSource` of a given type that's defined anywhere in the classpath. If someone wants to add a new permission source, it gets lumped in with all the others even though there's no way to tell that's going to happen. That's my concern and why I specifically don't think `AggregatingResourcePermissionProvider` belongs in the code base.

AbdulRahmanAlHamali: I see what you mean, but I still think an aggregating provider is a general-enough usecase, because `AggregatingResourcePermissionProvider` works great with `deny-all` models, that start by denying all privileges from users, and then adding them incrementally

plumpy: But what if we want to use a different model? What if I want to check in a `AllowByDefaultResourcePermissionProvider` that uses a new `ForbiddenResourcePermissionSource`. That provider is subtractive, rather than additive.

Is that something we'd want to do? I don't know! But as this stands there is _no way_ to check in a `ResourcePermissionSource` like that without breaking everyone who sets their permission provider to `aggregate`. You've tied our hands as to what sorts of `ResourcePermissionSources` are allowed, forever.

plumpy: I suppose the answer there is that we just have to define a different interface to use, other than `ResourcePermissionSource.`

hebah4change: I'm not sure the example you're giving provides an actual usecase.

If we think of it this way: `Sources` are sources from which we read permissions of a resource, while `Providers` are the ones who read all those permissions, and decide how to consolidate them. A `ForbiddenSource` is a source trying to play the role of a provider, and trying to dictate how sources other than itself should be interpreted, which doesn't make sense.

In general, I don't think a subtractive provider really works. I think to implement an allow-by-default strategy, you need a "priority" provider, which provides full permissions if no sources exist, and chooses the highest priority source (based on some rules) otherwise

AbdulRahmanAlHamali: I'm not sure the example you're giving provides an actual usecase.

If we think of it this way: Sources are sources from which we read permissions of a resource, while Providers are the ones who read all those permissions, and decide how to consolidate them. A ForbiddenSource is a source trying to play the role of a provider, and trying to dictate how sources other than itself should be interpreted, which doesn't make sense.

In general, I don't think a subtractive provider really works. I think to implement an allow-by-default strategy, you need a "priority" provider, which provides full permissions if no sources exist, and chooses the highest priority source (based on some rules) otherwise

PR #485, opened 2019-10-21
feature(authorization): Add creation access control

AbdulRahmanAlHamali:
**This implementation is still in draft state**. It achieves the following:
1. Add a configuration parameter called `restrictApplicationCreation`. If set to false (the default), this feature will be ignored.
2. Add a `CREATE` authorization.
3. Call the `ResourcePermissionProvider` immediately from `AuthorizeController`, because creation permissions are typically dynamic, and cannot be known before the application is submitted for creation. Thus, they cannot be held on the UserPermissions object.
4. Make `Front50ApplicationResourcePermission` source strip out any `CREATE` permissions provided on the source, because it doesn't make sense for users to add `CREATE` permissions on the resource they are creating, and it will basically work against any other authorization model we try to implement.
5.  Implementations of `ResourcePermissionProvider` and `ResourcePermissionSource` will be free to handle creation permissions the way they say fit.

Please let me know what you think. Let's try to release this in the Oct 28th release

### DIFF
--- a/fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatPermissionEvaluator.java
+++ b/fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatPermissionEvaluator.java
@@ -25,6 +25,7 @@ import com.netflix.spinnaker.fiat.model.UserPermission;
 import com.netflix.spinnaker.fiat.model.resources.Account;
 import com.netflix.spinnaker.fiat.model.resources.Authorizable;
 import com.netflix.spinnaker.fiat.model.resources.ResourceType;
+import com.netflix.spinnaker.kork.exceptions.IntegrationException;
 import com.netflix.spinnaker.security.AuthenticatedRequest;
 import com.netflix.spinnaker.security.User;
 import java.io.Serializable;
@@ -39,19 +40,18 @@ import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Function;
 import java.util.stream.Collectors;
 import javax.annotation.Nonnull;
-import javax.servlet.ServletException;
-import javax.servlet.http.HttpServletResponse;
 import lombok.extern.slf4j.Slf4j;
 import lombok.val;
 import org.apache.commons.lang3.StringUtils;
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.http.HttpStatus;
 import org.springframework.security.access.PermissionEvaluator;
 import org.springframework.security.core.Authentication;
 import org.springframework.security.core.context.SecurityContextHolder;
 import org.springframework.stereotype.Component;
 import org.springframework.util.backoff.BackOffExecution;
 import org.springframework.util.backoff.ExponentialBackOff;
-import retrofit.client.Response;
+import retrofit.RetrofitError;
 
 @Component
 @Slf4j
@@ -160,16 +160,24 @@ public class FiatPermissionEvaluator implements PermissionEvaluator {
       return AuthenticatedRequest.propagate(
               () -> {
                 return retryHandler.retry(
-                    "determine whether " + username + " can create resource" + resource,
+                    "determine whether " + username + " can create resource " + resource,
                     () -> {
-                      Response response = fiatService.canCreate(username, resourceType, resource);
-                      if (response.getStatus() == HttpServletResponse.SC_OK) {
+                      try {
+                        fiatService.canCreate(username, resourceType, resource);
                         return true;
-                      } else if (response.getStatus() == HttpServletResponse.SC_NOT_FOUND) {
-                        return false;
-                      } else {
-                        throw new ServletException(
-                            response.getStatus() + ": " + response.getReason());
+                      } catch (RetrofitError re) {
+                        boolean shouldRetry = true;
+                        if (re.getKind() == RetrofitError.Kind.HTTP) {
+                          switch (HttpStatus.valueOf(re.getResponse().getStatus())) {
+                            case NOT_FOUND:
+                              return false;
+                            case BAD_REQUEST:
+                              shouldRetry = false;
+                          }
+                        }
+                        IntegrationException ie = new IntegrationException(re);
+                        ie.setRetryable(shouldRetry);
+                        throw ie;
                       }
                     });
               })
--- a/fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatPermissionEvaluator.java
+++ b/fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatPermissionEvaluator.java
@@ -24,7 +24,6 @@ import com.netflix.spinnaker.fiat.model.Authorization;
 import com.netflix.spinnaker.fiat.model.UserPermission;
 import com.netflix.spinnaker.fiat.model.resources.Account;
 import com.netflix.spinnaker.fiat.model.resources.Authorizable;
-import com.netflix.spinnaker.fiat.model.resources.Resource;
 import com.netflix.spinnaker.fiat.model.resources.ResourceType;
 import com.netflix.spinnaker.security.AuthenticatedRequest;
 import com.netflix.spinnaker.security.User;
@@ -147,36 +146,23 @@ public class FiatPermissionEvaluator implements PermissionEvaluator {
   @Override
   public boolean hasPermission(
       Authentication authentication, Object resource, Object authorization) {
+    return false;
+  }
+
+  public boolean canCreate(String resourceType, Object resource) {
     if (!fiatStatus.isEnabled()) {
       return true;
     }
 
-    Resource r;
-    try {
-      if (resource == null) {
-        throw new IllegalArgumentException("Provided resource is null");
-      }
-      r = (Resource) resource;
-    } catch (ClassCastException e) {
-      throw new IllegalArgumentException(
-          "Provided resource is not an instance of " + Resource.class.getCanonicalName());
-    }
-
-    Authorization a = Authorization.valueOf(authorization.toString());
-    if (a != Authorization.CREATE) {
-      throw new IllegalArgumentException(
-          "This method should only be called for `CREATE`. For other operations, please call the other implementation");
-    }
-
-    String userName = getUsername(authentication);
+    String username = getUsername(SecurityContextHolder.getContext().getAuthentication());
 
     try {
       return AuthenticatedRequest.propagate(
               () -> {
                 return retryHandler.retry(
-                    "get whether " + userName + " can " + a.toString() + " resource " + resource,
+                    "get whether " + username + " can create resource" + resource,
                     () -> {
-                      Response response = fiatService.hasAuthorization(userName, a.toString(), r);
+                      Response response = fiatService.canCreate(username, resourceType, resource);
                       if (response.getStatus() == HttpServletResponse.SC_OK) {
                         return true;
                       } else if (response.getStatus() == HttpServletResponse.SC_NOT_FOUND) {
@@ -212,14 +198,14 @@ public class FiatPermissionEvaluator implements PermissionEvaluator {
     ResourceType r = ResourceType.parse(resourceType);
     Authorization a = null;
 
-    if (a == Authorization.CREATE) {
-      throw new IllegalArgumentException(
-          "This method should not be called for `CREATE`. Please call the other implementation");
-    }
-
     // Service accounts don't have read/write authorizations.
     if (r != ResourceType.SERVICE_ACCOUNT) {
       a = Authorization.valueOf(authorization.toString());
+    }
+
+    if (a == Authorization.CREATE) {
+      throw new IllegalArgumentException(
+          "This method should not be called for `CREATE`. Please call the other implementation");
     }
 
     if (r == ResourceType.APPLICATION && StringUtils.isNotEmpty(resourceName.toString())) {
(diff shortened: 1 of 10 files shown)


fiat-web/src/main/java/com/netflix/spinnaker/fiat/controllers/AuthorizeController.java
```diff
@@ -150,35 +147,54 @@ public AuthorizeController(
   }
```
AbdulRahmanAlHamali: I realize I should not modify the function below, and create a special one to handle `CREATE`, because the function below is name-based, while for `CREATE` I want to handle the whole object not just its name.
Nevertheless, I will keep this for now for you to look at, and will modify later.
It is also worth it to (in a different PR), make a breaking change that migrates all the usages to use the full resource instead of just the name.

fiat-web/src/main/java/com/netflix/spinnaker/fiat/controllers/AuthorizeController.java
```diff
@@ -150,35 +147,54 @@ public AuthorizeController(
...
       HttpServletResponse response)
       throws IOException {
     Authorization a = Authorization.valueOf(authorization.toUpperCase());
-    ResourceType r = ResourceType.parse(resourceType);
+    ResourceType rt = ResourceType.parse(resourceType);
     Set<Authorization> authorizations = new HashSet<>(0);
 
-    try {
-      switch (r) {
-        case ACCOUNT:
```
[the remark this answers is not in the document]
AbdulRahmanAlHamali: yup I actually never found any calls for it in other services

fiat-web/src/main/java/com/netflix/spinnaker/fiat/controllers/AuthorizeController.java
```diff
@@ -128,6 +123,22 @@ public AuthorizeController(
         .orElseThrow(NotFoundException::new);
   }
```
AbdulRahmanAlHamali: I added these below since we didn't have them

AbdulRahmanAlHamali:
@plumpy would you kindly provide feedback on this PR?

fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatPermissionEvaluator.java
```diff
@@ -143,7 +147,51 @@ private static RetryHandler buildRetryHandler(
   @Override
   public boolean hasPermission(
       Authentication authentication, Object resource, Object authorization) {
-    return false;
+    if (!fiatStatus.isEnabled()) {
+      return true;
+    }
+
+    Resource r;
```
cfieber: This is tricky, but I think we need to think through it a bit more

This code is actually running in-process in the fiat clients, and is annotated onto methods there. (For example in Front50 on a `getApplications` endpoint). In that case, the type is going to be that applications API model type, which isn't going to implement Resource.

Not sure I have a great suggestion immediately, thinking about it..

fiat-web/src/main/java/com/netflix/spinnaker/fiat/controllers/AuthorizeController.java
```diff
@@ -189,6 +200,64 @@ public void getUserAuthorization(
     response.setStatus(HttpServletResponse.SC_NOT_FOUND);
   }
 
+  @RequestMapping(value = "/{userId:.+}/{authorization:.+}", method = RequestMethod.POST)
+  public void hasAuthorization(
+      @PathVariable String userId,
+      @PathVariable String authorization,
+      @RequestBody @Nonnull Resource resource,
```
cfieber: Follow-on to the previous comment - we will need to see how we go from user domain object to Resource but this might need `@PathVariable ResourceType` and `@RequestBody Map<String, Object> resource` so we can handle the object conversion properly - currently the Resource hierarchy isn't set up for Jackson polymorphic deserialization.

fiat-web/src/main/java/com/netflix/spinnaker/fiat/controllers/AuthorizeController.java
```diff
@@ -189,6 +204,42 @@ public void getUserAuthorization(
...
+          HttpServletResponse.SC_BAD_REQUEST,
+          "Resource type " + resourceType + "does not support creation");
+      return;
+    }
+    List<String> userRoles =
+        getUserRoles(userId).stream().map(Role.View::getName).collect(Collectors.toList());
+
+    Resource r = objectMapper.convertValue(resource, rt.modelClass);
+
+    // can easily implement options other than APPLICATION, but it is not currently needed.
```
cfieber: +1 I'm happy with keeping the scope narrow to the existing use-case for now

fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatPermissionEvaluator.java
```diff
@@ -146,6 +149,37 @@ public boolean hasPermission(
...
+      return true;
+    }
+
+    String username = getUsername(SecurityContextHolder.getContext().getAuthentication());
+
+    try {
+      return AuthenticatedRequest.propagate(
+              () -> {
+                return retryHandler.retry(
+                    "get whether " + username + " can create resource" + resource,
```
cfieber:
```suggestion
                    "determine whether " + username + " can create resource " + resource,
```

fiat-roles/src/main/java/com/netflix/spinnaker/fiat/providers/Front50ApplicationResourcePermissionSource.java
```diff
@@ -53,6 +53,9 @@ public Permissions getPermissions(@Nonnull Application resource) {
       authorizations.put(Authorization.EXECUTE, authorizations.get(executeFallback));
     }
 
+    // CREATE permissions are not allowed on the resource level
```
cfieber:
```suggestion
    // CREATE permissions are not allowed on the resource level.
    //  At this point the resource already exists, so CREATE does
    //  not make sense.
```

fiat-web/src/main/java/com/netflix/spinnaker/fiat/controllers/AuthorizeController.java
cfieber: I'd prefer to just leave these out - we don't seem to have a need for them anywhere yet

fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatPermissionEvaluator.java
```diff
@@ -146,6 +149,37 @@ public boolean hasPermission(
...
+                return retryHandler.retry(
+                    "determine whether " + username + " can create resource" + resource,
+                    () -> {
+                      Response response = fiatService.canCreate(username, resourceType, resource);
+                      if (response.getStatus() == HttpServletResponse.SC_OK) {
+                        return true;
+                      } else if (response.getStatus() == HttpServletResponse.SC_NOT_FOUND) {
+                        return false;
+                      } else {
+                        throw new ServletException(
```
cfieber: lets pick something other than ServletException here. I think IntegrationException with setRetryable(true) would be more appropriate (we may want to smarten up retry handler in the future, so there is a preference to use subclasses of SpinnakerException wherever possible)

cfieber: should have spinnaker/kork#401 merged in to make that a litle less painful

fiat-web/src/main/java/com/netflix/spinnaker/fiat/controllers/AuthorizeController.java
```diff
@@ -189,6 +188,42 @@ public void getUserAuthorization(
...
+    if (applicationResourcePermissionProvider
```
cfieber: think we might need an isAdmin check here on user as well

fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatPermissionEvaluator.java
cfieber: will also not want to retry on BAD_REQUEST

fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatPermissionEvaluator.java
```diff
@@ -146,6 +149,37 @@ public boolean hasPermission(
...
+
+    String username = getUsername(SecurityContextHolder.getContext().getAuthentication());
+
+    try {
+      return AuthenticatedRequest.propagate(
+              () -> {
+                return retryHandler.retry(
+                    "determine whether " + username + " can create resource" + resource,
+                    () -> {
+                      Response response = fiatService.canCreate(username, resourceType, resource);
```
cfieber: The response handling bit in here is incorrect.

On a 4xx response from the server, retrofit throws a RetrofitException

This should looks something like:

```suggestion
                      try {
                        fiatService.canCreate(username, resourceType, resource);
                        return true;
                      } catch (RetrofitError re) {
                        boolean shouldRetry = true;
                        if (re.getKind() == RetrofitError.Kind.HTTP) {
                          switch (HttpStatus.valueOf(re.getResponse().getStatus())) {
                            case NOT_FOUND:
                              return false;

                            case BAD_REQUEST:
                              shouldRetry = false;
                              break;
                          }
                        }
                        IntegrationException ie = new IntegrationException(re);
                        ie.setRetryable(shouldRetry);
                        throw ie;
                      }
```

fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatPermissionEvaluator.java
cfieber: see https://github.com/spinnaker/fiat/pull/485/files#r337840808

fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatPermissionEvaluator.java
```diff
@@ -146,6 +149,37 @@ public boolean hasPermission(
...
+      return true;
+    }
+
+    String username = getUsername(SecurityContextHolder.getContext().getAuthentication());
+
+    try {
+      return AuthenticatedRequest.propagate(
+              () -> {
+                return retryHandler.retry(
+                    "determine whether " + username + " can create resource" + resource,
```
cfieber:
```suggestion
                    "determine whether " + username + " can create resource " + resource,
```

fiat-web/src/main/java/com/netflix/spinnaker/fiat/controllers/AuthorizeController.java
```diff
@@ -189,6 +188,42 @@ public void getUserAuthorization(
...
+
+    if (!configProps.isRestrictApplicationCreation()) {
+      response.setStatus(HttpServletResponse.SC_OK);
+      return;
+    }
+
+    if (rt != ResourceType.APPLICATION) {
+      response.sendError(
+          HttpServletResponse.SC_BAD_REQUEST,
+          "Resource type " + resourceType + "does not support creation");
```
cfieber:
```suggestion
          "Resource type " + resourceType + " does not support creation");
```

fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatService.java
```diff
@@ -50,6 +50,18 @@ Response hasAuthorization(
...
       @Path("authorization") String authorization);
 
+  /**
+   * @param userId The username of the user
+   * @param resourceType The type of the resource
+   * @param resource The resource to check
+   * @return 200 if the user can create the specified resource, 404 otherwise
+   */
+  @POST("/authorize/{userId}/{resourceType}")
+  Response canCreate(
```
cfieber:
```suggestion
  void canCreate(
```
this can be void instead of Response - we only care about the Response if we got a RetrofitError and in that case we can get it from there

fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatPermissionEvaluator.java
AbdulRahmanAlHamali: The retry handler doesn't look like it cares about the `retryable` property. Or is this for the future?

AbdulRahmanAlHamali:
> other than the RetrofitError handling and a couple minor spacing tweaks this looks good
> 
> I was able to build locally and integrate into front50 and add a ResourcePermissionSource that added a CREATE authorization to the application and see it properly enforced.
> 
> I am not proposing changing it here (lets revisit in the next PR that tries to add something supplying CREATE permissions for Applications), but I do suspect we will need to enhance the ResourcePermissionProvider/Source model to indicate that we are asking for Permissions for a new resource - in my contrived ResourcePermissionSource for testing I felt like it would be hard to write a legitimate ResourcePermissionSource to add CREATE and also support the case of not causing a resource to go from isRestricted=false to isRestricted=true.
> 
> But, let's dig into that in the next PR..

Alright, I have addressed your comments :)

Regarding distinguishing new resources, I am trying in this PR to get things done without distinguishing, but it might indeed be too hard. I'll start the next PR for the PrefixSource, which will make things clearer for us.

PR #490, opened 2019-10-23
feat(authorization): implement application prefix permission source

AbdulRahmanAlHamali:
**This is still a draft**
This PR allows users to provide permissions to applications using name prefixes specified like:
- "*"
- "abc*"
- etc.


fiat-roles/src/main/java/com/netflix/spinnaker/fiat/providers/ApplicationPrefixPermissionSource.java
```diff
@@ -0,0 +1,53 @@
...
+
+package com.netflix.spinnaker.fiat.providers;
+
+import com.netflix.spinnaker.fiat.model.resources.Application;
+import com.netflix.spinnaker.fiat.model.resources.Permissions;
+import javax.annotation.Nonnull;
+import lombok.Data;
+
+@Data
+public class ApplicationPrefixPermissionSource implements ResourcePermissionSource<Application> {
```
cfieber: I would make this a singleton component that accepts the configuration map of prefix -> permissions

in getPermissions iterate over the map finding matching keys (I guess all matching prefixes considering there could be overlap) and build up the resulting Permissions object from that

This could be made generic (`PrefixResourcePermissionSource<T extends Resource> implements ResourcePermissionSource<T>`) since there really isn't anything Application specific in here and I could definitely see legitimate uses for this on Account/BuildService as well. (In DefaultResourcePermissionConfig you could still just create the one for `T = Application` for now)

If you think it is more appropriate to create an instance of this per prefix mapping within a type to allow fancier ResourcePermissionProvider aggregating, it should still be done via constructor args rather than `@Data` and a setter.  In that case you'll need to update the DefaultResourcePermissionConfig to create each instance and register it with the `configurableApplicationContext.getBeanFactory().registerSingleton("name", prefixSpecificSource)` rather than returning a List.

fiat-web/src/main/java/com/netflix/spinnaker/fiat/config/DefaultResourcePermissionConfig.java
```diff
@@ -63,6 +70,15 @@
         fiatServerConfigurationProperties.getExecuteFallback());
   }
 
+  @Bean
+  @ConditionalOnExpression(
+      "T(org.springframework.utils.CollectionUtils).isNotEmpty(${auth.permissions.source.application.prefix})")
+  List<ResourcePermissionSource<Application>> applicationPrefixPermissionSource(
```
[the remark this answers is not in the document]
cfieber: see the comment in ApplicationPrefixPermissionSource above - this should either return one ResourcePermissionSource that can handle all the supplied prefixes, or create one instance per prefix but then register those with the applicationContext

fiat-roles/src/main/java/com/netflix/spinnaker/fiat/providers/ApplicationPrefixPermissionSource.java
AbdulRahmanAlHamali: I apologize for the rushed implementation. I wanted to submit something quick for you to look at before I went home.

I will fix the way permissions are read from the config. But regarding having just one source that contains all the prefixes, I preferred to have one source per prefix, because, as you mentioned, it will allow to resolve them using any desired provider. If I put their resolution logic inside the source, I will be replicating the work of the `AggregatingResourcePermissionProvider`. Also, it might be desired to resolve them in different ways. For example, more specific might be desired to override less specific, which will be harder to do if I rule what resolution logic should be supported.

cfieber: agreed, I think multiple sources is the right way to go

AbdulRahmanAlHamali:
I have reimplemented this as a single source. Implementing it as multiple sources and getting the beans to load correctly in Spring was very hacky and error-prone. So now it is a single source with multiple prefixes, and a resolution strategy which could either be `aggregate` or `most_specific`.

I also added the ability to add a full name, in which case I do exact matching. This allows prefix source to function on its own without any other sources if the admin desires.

fiat-roles/src/main/java/com/netflix/spinnaker/fiat/providers/ResourcePrefixPermissionSource.java
```diff
@@ -0,0 +1,122 @@
...
+public class ResourcePrefixPermissionSource<T extends Resource.AccessControlled>
+    implements ResourcePermissionSource<T> {
+
+  public enum ResolutionStrategy {
+    AGGREGATE,
+    MOST_SPECIFIC
+  }
+
+  private List<PrefixEntry<T>> prefixes;
+  private ResolutionStrategy resolutionStrategy;
```
cfieber: would it make sense to have a default here? (AGGREGATE perhaps?)

PR #500, opened 2019-11-04
fix(unrestricted): load permissions for unrestricted roles

cfieber:
The unrestricted user can have associated roles via
`UserRolesProvider.loadUnrestrictedRoles`, however this would
always short-circuit evaluation of permissions for those roles.

This resulted in the case where on the first authorization
(`POST /roles/:userId`) the user would not have access to any
resources that were granted solely via the unrestricted role.
However in a subsequent authentication - the users permissions
would get merged with an existing user entry and the unrestricted
roles would show up as if they were directly granted to the user
(due to thge way the `UserPermissions.merge` happens in the
`RedisPermissionRepository`) and eventually the resources protected
by unrestricted roles would show up.


ajordens:
Ahh this was a neat one -- good find.

Think it warrants a test case?

cfieber:
yeah probably so, will add one

PR #502, opened 2019-11-05
fix(authorization): canCreate should not return void

AbdulRahmanAlHamali:
Retrofit methods cannot have a `void` return type. We either need to provide a return type, or a callback. This fixes the following exception:
```
java.lang.IllegalArgumentException: FiatService.canCreate: Must have either a return type or Callback as last argument.
```

I also added `canCreate` to the path, because it makes more sense.

### DIFF
--- a/fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatService.java
+++ b/fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatService.java
@@ -58,8 +58,8 @@ public interface FiatService {
    * @param resourceType The type of the resource
    * @param resource The resource to check
    */
-  @POST("/authorize/{userId}/{resourceType}")
-  void canCreate(
+  @POST("/authorize/canCreate/{userId}/{resourceType}")
+  Response canCreate(
       @Path("userId") String userId,
       @Path("resourceType") String resourceType,
       @Body Object resource);
--- a/fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatService.java
+++ b/fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatService.java
@@ -58,7 +58,7 @@ public interface FiatService {
    * @param resourceType The type of the resource
    * @param resource The resource to check
    */
-  @POST("/authorize/canCreate/{userId}/{resourceType}")
+  @POST("/authorize/{userId}/{resourceType}/create")
   Response canCreate(
       @Path("userId") String userId,
       @Path("resourceType") String resourceType,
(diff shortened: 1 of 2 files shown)


fiat-web/src/main/java/com/netflix/spinnaker/fiat/controllers/AuthorizeController.java
```diff
@@ -188,7 +188,7 @@ public void getUserAuthorization(
     response.setStatus(HttpServletResponse.SC_NOT_FOUND);
   }
 
-  @RequestMapping(value = "/{userId:.+}/{resourceType:.+}", method = RequestMethod.POST)
+  @RequestMapping(value = "canCreate/{userId:.+}/{resourceType:.+}", method = RequestMethod.POST)
```
ajordens: This is a backwards incompatible change if something/someone is already calling this endpoint?

I don't believe `canCreate` belongs in the path personally ... 

What about overloading the existing `/authorize` with a `@POST` that takes the payload -- in this case I'd expect that we only support `authorization == create`?

```
@POST("/authorize/{userId}/{resourceType}/{resourceName}/{authorization}")
```

Thoughts on this @cfieber?

ajordens: 👍

fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatService.java
```diff
@@ -58,8 +58,8 @@ Response hasAuthorization(
    * @param resourceType The type of the resource
    * @param resource The resource to check
    */
-  @POST("/authorize/{userId}/{resourceType}")
-  void canCreate(
+  @POST("/authorize/canCreate/{userId}/{resourceType}")
```
cfieber:
```suggestion
  @POST("/authorize/{userId}/{resourceType}/create")
```

fiat-web/src/main/java/com/netflix/spinnaker/fiat/controllers/AuthorizeController.java
cfieber:
```suggestion
  @RequestMapping(value = "/{userId:.+}/{resourceType:.+}/create", method = RequestMethod.POST)
```

AbdulRahmanAlHamali: Sounds good, I applied the suggestions

cfieber:
I was wrong about not changing the URI - this is breaking since we always do always call fiat on CREATE (and the enable/disable flag is on the fiat side).

We should leave the URI as is, or if we want to migrate it to the new form we should maintain a request mapping in fiat for the old form for some period of time - otherwise we are stuck in the situation of breaking front50 until we deploy new fiat (and then breaking old front50 until we deploy new).

AbdulRahmanAlHamali:
The thing is, it is already broken in its current state, so if we sneak that change in, we won't break something that works

cfieber:
Ah you are right - good point.  Gonna :shipit:

cfieber:

PR #508, opened 2019-11-15
fix(roles): file-based roles fail when the user is not provided in the file

AbdulRahmanAlHamali:
When calling `loadRoles` of file-based roles. I am faced with a `NullPointerException` if the user is not defined in the file. I fixed this to return an empty list instead.

This is a valid use-case for implementations that use external roles (SAML for example), but want to provide specific roles for some users.


AbdulRahmanAlHamali:
Thank you for the merge! Can we cherry-pick this into 1.17 as well please?

AbdulRahmanAlHamali:

PR #513, opened 2019-11-22
fix(serviceAccount): Filter non-valid roles when converting to UserPermission

xavileon:
Roles can't be empty. This leads to unexpected behaviour. However, we
were allowing creating empty roles like "" or "   " on the pipeline
triggers (via API), which made service users to contain invalid roles
and thus failing on every sync request that tries to map roles to accounts.
This rendered FIAT unusable to get permissions and subsequently not
allowing any authorization operation.

This patch sanitizes the input on ServiceAccounts so we make sure
that the roles considered are valid.


xavileon:

PR #519, opened 2019-12-04
fix(permissionSource): Set order (lowest precedence) on default permission sources

jonsie:
Additionally, rename Front50ApplicationResourcePermissionSource to ApplicationResourcePermissionSource and adjust config accordingly.

This way, if default permission source objects contain logic then they will operate after any logic contained in an aggregate permission source (if order is specified correctly).

Specifically, we want to process `EXECUTE` permissions last in the chain of aggregate permission sources since if a resource is missing `EXECUTE` permissions it falls back to the configured permission type (defaults to `READ`).

### DIFF
--- a/fiat-roles/src/main/java/com/netflix/spinnaker/fiat/providers/ApplicationResourcePermissionSource.java
+++ b/fiat-roles/src/main/java/com/netflix/spinnaker/fiat/providers/ApplicationResourcePermissionSource.java
@@ -27,12 +27,12 @@ import java.util.List;
 import java.util.Map;
 import javax.annotation.Nonnull;
 
-public final class Front50ApplicationResourcePermissionSource
+public final class ApplicationResourcePermissionSource
     implements ResourcePermissionSource<Application> {
 
   private final Authorization executeFallback;
 
-  public Front50ApplicationResourcePermissionSource(Authorization executeFallback) {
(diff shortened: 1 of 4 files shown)


fiat-roles/src/main/java/com/netflix/spinnaker/fiat/providers/ApplicationResourcePermissionSource.java
```diff
@@ -27,12 +27,12 @@
 import java.util.Map;
 import javax.annotation.Nonnull;
 
-public final class Front50ApplicationResourcePermissionSource
+public final class ApplicationResourcePermissionSource
```
robzienert: nit: No disagreement here, but interested to know rationale.

jonsie: Essentially, this class contains a bit of logic around `EXECUTE` and `CREATE` permissions that we want to apply by default but these are not specific to processing front50 permissions - they could be any permission source.

jonsie: With that I should say, the impotence for this change is essentially so that we always process resource `EXECUTE` permissions last (since it falls back to `READ` if not set).

PR #526, opened 2019-12-10
fix(log): Add request headers to access denied exception handler log

jonsie:


fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatAccessDeniedExceptionHandler.java
```diff
@@ -73,6 +78,18 @@ private String authorizationFailureMessage(
     return sj.toString();
   }
 
+  private Map<String, String> requestHeaders(HttpServletRequest request) {
+    Map<String, String> headers = new HashMap<>();
+
+    while (request.getHeaderNames() != null && request.getHeaderNames().hasMoreElements()) {
+      String headerKey = request.getHeaderNames().nextElement();
```
marchello2000: won't the enumeration be reset every time you call `getHeaderNames()` (or give you a new enumeration in reset state?)

jonsie: Yes, dumb.

fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatAccessDeniedExceptionHandler.java
```diff
@@ -73,6 +79,20 @@ private String authorizationFailureMessage(
     return sj.toString();
   }
 
+  private Map<String, String> requestHeaders(HttpServletRequest request) {
+    Map<String, String> headers = new HashMap<>();
+
+    if (request.getHeaderNames() != null) {
+      for (Enumeration<?> e = request.getHeaderNames(); e.hasMoreElements(); ) {
```
marchello2000: nit: `e` -> `headerNames`

jonsie: I actually just changed that to `h`

PR #532, opened 2019-12-13
chore(permissions): Add unpack() public method to Permissions

jonsie:
Yea or nay?


fiat-core/src/main/java/com/netflix/spinnaker/fiat/model/resources/Permissions.java
```diff
@@ -100,6 +103,10 @@ public boolean isAuthorized(Set<Role> userRoles) {
     return permissions.getOrDefault(a, Collections.emptyList());
```
robzienert:
```suggestion
    return permissions.getOrDefault(a, new ArrayList<>());
```

PR #545, opened 2020-01-20
feat(chaos-monkey): Add ChaosMonkeyResourcePermissionSource and switch fiat to pulling list of applications vs. list of application permissions from front50

jonsie:
This is riffing off an idea @cfieber had regarding using Fiat as a source of permission policies.

Listing all Applications from front50 (as opposed to listing all application permissions) allows us to use application configuration to apply a permissions policy.  In this case, add the chaos monkey roles to READ and WRITE authorizations if the application config has chaos monkey enabled.  

Logically, this is far simpler than what currently exists in Front50 (see https://github.com/spinnaker/front50/tree/master/front50-core/src/main/groovy/com/netflix/spinnaker/front50/listeners) as, since the role is not stored in the application configs permissions in front50, there is far less logic that needs to be applied to determine when to add/remove the role so as not to lock people out of their applications.

### DIFF
--- a/fiat-roles/src/main/java/com/netflix/spinnaker/fiat/providers/ChaosMonkeyResourcePermissionSource.java
+++ b/fiat-roles/src/main/java/com/netflix/spinnaker/fiat/providers/ChaosMonkeyResourcePermissionSource.java
@@ -1,0 +1,67 @@
+/*
+ * Copyright 2020 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.fiat.providers;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.netflix.spinnaker.fiat.model.Authorization;
+import com.netflix.spinnaker.fiat.model.resources.Application;
+import com.netflix.spinnaker.fiat.model.resources.Permissions;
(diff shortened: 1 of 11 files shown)


fiat-roles/src/main/java/com/netflix/spinnaker/fiat/providers/ChaosMonkeyResourcePermissionSource.java
```diff
@@ -0,0 +1,67 @@
...
+  @Override
+  public Permissions getPermissions(@Nonnull T resource) {
+    Permissions.Builder builder = new Permissions.Builder();
+    Permissions permissions = resource.getPermissions();
+
+    if (permissions.isRestricted()) {
+      if (resource instanceof Application) {
+        Application application = (Application) resource;
+        if (isChaosMonkeyEnabled(application)) {
+          builder.add(Authorization.READ, roles).add(Authorization.WRITE, roles).build();
```
jonsie: Since the role is not ever stored on the application config, there is no "remove chaos monkey when" logic here - much simpler.

robzienert: Like this much more!

fiat-roles/src/main/java/com/netflix/spinnaker/fiat/providers/ChaosMonkeyResourcePermissionSource.java
```diff
@@ -0,0 +1,67 @@
...
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.netflix.spinnaker.fiat.model.Authorization;
+import com.netflix.spinnaker.fiat.model.resources.Application;
+import com.netflix.spinnaker.fiat.model.resources.Permissions;
+import com.netflix.spinnaker.fiat.model.resources.Resource;
+import java.util.List;
+import javax.annotation.Nonnull;
+
+public class ChaosMonkeyResourcePermissionSource<T extends Resource.AccessControlled>
```
cfieber: I think you could make this ChaosMonkeyResourcePermissionSource<Application> implements ResourcePermissionSource<Application> and get rid of the type check in getPermissions

robzienert:
ping - do we wanna promote from draft and do the thing?

jonsie:
Yeah, I'll get this moved over to a real PR today/tomorrow - just want to take one more look at the configuration before I do it.

PR #560, opened 2020-02-12
chore(vulnerabilities): resolve CVEs and other security issues

asgard-r:
- remove CVE-2019-10173, bump xstream dependency
- remove CVE-2018-18893, bump jinjava dependency

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -74,6 +74,7 @@ allprojects {
   configurations.all {
     resolutionStrategy {
       force group: 'com.thoughtworks.xstream', name: 'xstream', version: '1.4.+' // remove CVE-2019-10173, bump xstream dependency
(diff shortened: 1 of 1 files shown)


ajordens:
It's unclear to me why we have a `jinjava` dependency in the first place ... it looks to be coming from `kork-security` but I can't see a reason why it's there either.

Rather than doing force upgrades, can we just remove it (and presumably only use it in the limited number of spots it's needed).

It also looks like `xstream` is coming from `eureka-client` ... it's unclear to me if we can just blindly upgrade that or if we should look to pulling a new `eureka-client`. FWIW the `eureka-client` master branch is using 1.4.11.

robzienert:
Closing. Eureka was updated and `jinjava` should be handled separately in `kork-security`.

PR #564, opened 2020-02-26
fix(applications): Do not apply permissions when building the internal model of apps

PerGon:
In this PR https://github.com/spinnaker/fiat/pull/545 there was a change in the way fiat calls front50. From using the unprotected endpoint `front50/permissions/applications` to use the protected `front50/v2/applications`.

For all applications returned by `front50/v2/applications`, the @PostFilter in that endpoint, will call fiat for all applications it will return to check for permissions. Which makes calling `front50/v2/applications` really slow. Fiat is just building it's internal model of permissions, so it can fetch all apps without restrictions.


jonsie:
Not sure why this is still here and open since I merged it.

PR #573, opened 2020-03-04
perf(auth/ldap): Use group to user mapping to role sync

kowshikRoy:
Currently fiat sync LDAP roles by each users synchronously by making API call
to LDAP server. As the number of users grow, this could cause performace
issue.
Ldap server maintains the Group -> User mapping denoting which users are
the part of particular group. So, instead of fetching roles for each users,
we can fetch the whole Group->User mapping in one API call and perform
the role sync.

Trade-off:
* As the response of LDAP will huge compared to previous approach.
  This could put load on network
* For small users, the previous approach could be faster. That's why we
  encourage people to define a threshold after which the group call
  should be made.

Performance Improvement: From 16 mins -> 14sec.
Linked issue:https://github.com/spinnaker/spinnaker/issues/5508


fiat-ldap/src/main/java/com/netflix/spinnaker/fiat/config/LdapConfig.java
```diff
@@ -57,5 +57,9 @@ SpringSecurityLdapTemplate springSecurityLdapTemplate() throws Exception {
     String userSearchFilter;
     String groupSearchFilter = "(uniqueMember={0})";
     String groupRoleAttributes = "cn";
+
+    int thresholdToUseGroupFilter = 1000;
+    String groupsFilter = "";
```
evansb: Can you describe the case where this might be different from `groupSearchFilter`? It seems that we can simply use the same `groupSearchFilter`.

fiat-ldap/src/main/java/com/netflix/spinnaker/fiat/config/LdapConfig.java
```diff
@@ -57,5 +57,9 @@ SpringSecurityLdapTemplate springSecurityLdapTemplate() throws Exception {
...
+    String groupsUserAttributes = "member";
```
evansb: Nit: Standardise the name with `groupsRoleAttributes`, something like `groupUserAttributes`?

fiat-ldap/src/main/java/com/netflix/spinnaker/fiat/roles/ldap/LdapUserRolesProvider.java
```diff
@@ -93,12 +99,55 @@
...
+    if (users.size() > configProps.getThresholdToUseGroupFilter()
+        && StringUtils.isNotEmpty(configProps.getGroupsFilter())) {
+      Map<String, Collection<Role>> userRolesMap =
+          users.stream()
+              .collect(Collectors.toMap(ExternalUser::getId, ExternalUser::getExternalRoles));
+      Set<String> userIds = userRolesMap.keySet();
+      ldapTemplate
+          .search(
+              configProps.getGroupSearchBase(), configProps.getGroupsFilter(), new GroupToUsers())
+          .forEach(
```
evansb: I'm not with IDE in my hand right now but can we reduce the indentation level by doing something like

```
.flatMap(entry -> entry.getValue())
.filter(userIds::contains)
.forEach(...)
```

fiat-ldap/src/main/java/com/netflix/spinnaker/fiat/roles/ldap/LdapUserRolesProvider.java
```diff
@@ -93,12 +99,55 @@
         .collect(Collectors.toList());
   }
 
+  private class GroupToUsers implements AttributesMapper<Pair<String, List<String>>> {
+    public Pair<String, List<String>> mapFromAttributes(Attributes attrs) throws NamingException {
+      String group = attrs.get(configProps.getGroupRoleAttributes()).get().toString();
+      NamingEnumeration<?> members = attrs.get(configProps.getGroupsUserAttributes()).getAll();
+      List<String> member = new ArrayList<>();
+      while (members.hasMore()) {
```
evansb: You can also use `for`
```
for (NamingEnumeration<?> members = attrs.get(configProps.getGroupsUserAttributes()).getAll(); members.hasMore()) {

}
```

fiat-ldap/src/main/java/com/netflix/spinnaker/fiat/roles/ldap/LdapUserRolesProvider.java
```diff
@@ -93,12 +99,55 @@
...
+        try {
+          String user =
+              Arrays.stream(configProps.getUserDnPattern().parse(members.next().toString()))
```
evansb: why can't we simply cast `Object[] [0]` to String here? seems to yield cleaner code

fiat-ldap/src/main/java/com/netflix/spinnaker/fiat/config/LdapConfig.java
evansb: Also put this below `groupRoleAttributes`

fiat-ldap/src/main/java/com/netflix/spinnaker/fiat/config/LdapConfig.java
kowshikRoy: `groupSearchFilter` is used by when fiat queries LDAP for each users individually. 
This commit doesn't replace the previous behaviour, but gives a way to make it faster if the user choose to adopt this change.

fiat-ldap/src/main/java/com/netflix/spinnaker/fiat/roles/ldap/LdapUserRolesProvider.java
kowshikRoy: Modified.

fiat-ldap/src/main/java/com/netflix/spinnaker/fiat/config/LdapConfig.java
evansb: sorry, meant to say `groupSearchBase` here

evansb: f2f discussion: I see why this may be useful but it seems more apt to be a separate PR. the non-optimised path will have to consider `groupFilter` after applying `groupSearchFilter` if it were introduced.

kowshikRoy:
@robzienert @ajordens @cfieber 
Please review this changes.

cfieber:
this looks fine, but I'd like to see a test that exercises this behaviour in LdapUserRolesProviderSpec?

kowshikRoy:
@cfieber added tests. Thanks.
Also, working on adding support for configuration in halyard

PR #576, opened 2020-03-04
fix(decode): Remove duplicate url decoder in ControllerSupport

jonsie:
A call to decode the input is not necessary since Spring decodes the path variable input by default.  See Spring's RequestMappingHandlerMapping and UrlPathHelper.  For example, a call to
fiat `/authorize/some%2Buser%40example.com` is properly decoded as `some+user@example.com` by Spring, but then decoded again via `ControllerSupport` -- the result is `some user@example.com`.  This can, in certain scenarios, cause authorization lookups to return a 404.


ajordens:
Thanks for chasing this one down!

PR #581, opened 2020-03-09
feat(build): Add GitHub Action workflows for master and release builds

jonsie:


jonsie:
I can add a candidate build using the `prereleased` event later, but I want to verify the basics first.

.github/workflows/pr.yml
```diff
@@ -1,16 +1,11 @@
-name: Fiat CI
```
ajordens: The name change here might bork the existing branch conditions we've got setup.

PR #601, opened 2020-03-13
chore(build): Bump spinnaker-gradle-project version

jonsie:
Also took the opportunity to rename some things in github action workflows.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,6 +1,6 @@
 #Fri Mar 06 02:45:42 UTC 2020
 enablePublishing=false
-spinnakerGradleVersion=7.1.1
+spinnakerGradleVersion=7.7.0
 korkVersion=7.25.1
 org.gradle.parallel=true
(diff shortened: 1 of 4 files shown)


gradle.properties
```diff
@@ -1,6 +1,6 @@
 #Fri Mar 06 02:45:42 UTC 2020
 enablePublishing=false
-spinnakerGradleVersion=7.1.1
+spinnakerGradleVersion=7.7.0
```
jonsie: @cfieber You reverted the last bump to both `spinnakerGradleVersion` and the project Gradle version.  Do you know if that publishing issue was due to the Gradle version bump or this dependency bump?

cfieber: it was due to me trying to get us onto gradle 6.x at the same time, which didn't fly

cfieber: if you can run a `publishToMavenLocal` I think everything should be happy

PR #602, opened 2020-03-13
fix(mergify): Set mergify to look at GA build status

jonsie:


.github/workflows/build.yml
```diff
@@ -7,7 +7,7 @@ on:
     - version-*
 
 jobs:
-  build:
+  branch-build:
```
jonsie: Status in GitHub is just the name of the GitHub Action job - seems safe to just have one job called `build` (in the `pr.yml` file) and rename this to `branch-build`.

PR #603, opened 2020-03-13
fix(build): Fetch all tags before proceeding with release build

jonsie:


jonsie:
I wonder if this is failing to merge because I'm changing the job that mergify depends on.  I'll manually merge and test with a different PR.

PR #606, opened 2020-03-13
refactor(rpc): reworks rpc to clouddriver/igor/front50

cfieber:
introduces a DataLoader that polls a remote collection and keeps a cache of it, and adds implementations
for the various collections from the services we aggregate to build up user permissions.

uses resilience4j instead of (the only one place we used it) hystrix to provide circuitbreaker/retry/fallback
on that collection.


cfieber:
this is kind of boilerplatey with the subclasses but that was necessary to be able to use annotation driven resilience4j and have a different retry/circuitbreaker instance for each remote service.. in all I'd rather do it this way than programmatically building up the circuit breaker and retry because the AOP stuff does lots of heavy lifting (more than the trivial subclassing here does)

fiat-roles/src/main/java/com/netflix/spinnaker/fiat/providers/internal/ClouddriverApplicationLoader.java
```diff
@@ -0,0 +1,27 @@
...
+
+package com.netflix.spinnaker.fiat.providers.internal;
+
+import com.netflix.spinnaker.fiat.model.resources.Application;
+import com.netflix.spinnaker.fiat.providers.ProviderHealthTracker;
+
+public class ClouddriverApplicationLoader extends ClouddriverDataLoader<Application> {
+  public ClouddriverApplicationLoader(
+      ProviderHealthTracker healthTracker, ClouddriverApi clouddriverApi) {
+    super(healthTracker, clouddriverApi::getApplications);
```
cfieber: boilerplateyness all for which Supplier<List<T>> to use

fiat-roles/src/main/java/com/netflix/spinnaker/fiat/providers/internal/ClouddriverDataLoader.java
```diff
@@ -0,0 +1,49 @@
...
+ */
+public class ClouddriverDataLoader<T> extends DataLoader<T> {
+  public ClouddriverDataLoader(
+      ProviderHealthTracker healthTracker, Supplier<List<T>> loadingFunction) {
+    super(healthTracker, loadingFunction);
+  }
+
+  @Override
+  @CircuitBreaker(name = "clouddriver", fallbackMethod = "getFallback")
+  @Retry(name = "clouddriver", fallbackMethod = "getFallback")
```
cfieber: boilerplateyness all for which circuitbreaker/retry name to use, and the fixedDelayString on the cache refresher

fiat-roles/src/main/java/com/netflix/spinnaker/fiat/providers/internal/DataLoader.java
```diff
@@ -0,0 +1,89 @@
...
+    List<T> data = allowAnonymous(loadingFunction::get);
+    return Collections.unmodifiableList(data);
+  }
+
+  private void logFallback(String resource, Throwable cause) {
+    String message = cause != null ? "Cause: " + cause.getMessage() : "";
+    log.info("Falling back to {} cache. {}", resource, message);
+  }
+
+  @Scheduled(fixedDelayString = "${fiat.front50-refresh-ms:30000}")
```
cfieber: this annotation should go away

cfieber:
thorough testing by way of turning off wifi and observing the fallbacks being hit...

fiat-roles/src/main/java/com/netflix/spinnaker/fiat/providers/internal/Front50Api.java
```diff
@@ -26,9 +26,17 @@
...
 
+  /**
+   * @deprecated for fiat's usage this is always going to be called with restricted = false, use the
+   *     no arg method instead which has the same behavior.
+   */
   @GET("/v2/applications")
+  @Deprecated
   List<Application> getAllApplications(@Query("restricted") boolean restricted);
 
+  @GET("/v2/applications?restricted=false")
```
jonsie: Nice.

fiat-roles/src/main/resources/resilience4j-defaults.properties
```diff
@@ -0,0 +1,49 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+#
+# Circuit breaker defaults
```
jonsie: Defining 3rd party defaults this way is a big 👍 from me.

fiat-roles/src/main/java/com/netflix/spinnaker/fiat/providers/internal/DataLoader.java
```diff
@@ -0,0 +1,87 @@
...
+
+  private final Supplier<List<T>> loadingFunction;
+  private final ProviderHealthTracker healthTracker;
+
+  private AtomicReference<List<T>> cache = new AtomicReference<>();
+
+  public DataLoader(ProviderHealthTracker healthTracker, Supplier<List<T>> loadingFunction) {
+    this.healthTracker = healthTracker;
+    this.loadingFunction = loadingFunction;
+    getData();
```
jonsie: I wonder if there is a way to call this after Spring has instantiated beans, like on a Spring event?

cfieber: I changed it to fire on ContextRefreshedEvent

PR #613, opened 2020-03-20
feat(ops): operational improvements

cfieber:
additional metrics around UserRolesSyncer timing
additional metrics on ResourceProvidersHealthIndicators
update ResourceProviderHealthIndicators to not go out of service
 once a resource has successfully loaded

### DIFF
--- a/fiat-roles/src/main/java/com/netflix/spinnaker/fiat/config/ResourceProvidersHealthIndicator.java
+++ b/fiat-roles/src/main/java/com/netflix/spinnaker/fiat/config/ResourceProvidersHealthIndicator.java
@@ -16,6 +16,8 @@
 
 package com.netflix.spinnaker.fiat.config;
 
+import com.netflix.spectator.api.Id;
+import com.netflix.spectator.api.Registry;
 import com.netflix.spinnaker.fiat.providers.HealthTrackable;
 import java.util.List;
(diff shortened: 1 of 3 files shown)


fiat-roles/src/main/java/com/netflix/spinnaker/fiat/config/ResourceProvidersHealthIndicator.java
```diff
@@ -34,27 +36,51 @@
...
+
   @Override
   protected void doHealthCheck(Health.Builder builder) throws Exception {
     boolean isDown = false;
     for (HealthTrackable provider : providers) {
       builder.withDetail(
           provider.getClass().getSimpleName(), provider.getHealthTracker().getHealthView());
       isDown = isDown || !provider.getHealthTracker().isProviderHealthy();
+      registry
+          .gauge(id("healthy", provider))
```
dreynaud: doesn't this create a new gauge with the same id on every `doHealthcheck` call? I think this is a legitimate use case for `PolledMeter` instead

cfieber: you can safely do `registry.gauge("foo).set(value)` repeatedly - it doesn't allocate a new object each time, just tracks the one measurement per id

PR #615, opened 2020-03-20
chore(build): Break fiat bump dependencies out into separate workflow

jonsie:
Exact same as https://github.com/spinnaker/kork/pull/567


jonsie:
Yea looks like the indentation was off on that one, this one is OK.  I'll wait for that release build to run before I merge this though.

PR #618, opened 2020-03-25
fix(permissions): update order annotation on sources

cfieber:
ensure we read the base resource permission first so other permission sources that augment Permissions
are processed afterword

also adds a scheduled refresh to the BaseResourceProvider cache to ensure a user request doesn't hit
a cache expiration and have to wait for the cache reload

### DIFF
--- a/fiat-roles/src/main/java/com/netflix/spinnaker/fiat/providers/BaseResourceProvider.java
+++ b/fiat-roles/src/main/java/com/netflix/spinnaker/fiat/providers/BaseResourceProvider.java
@@ -30,6 +30,7 @@ import java.util.stream.Collectors;
 import lombok.NonNull;
 import lombok.extern.slf4j.Slf4j;
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.scheduling.annotation.Scheduled;
 
 @Slf4j
 public abstract class BaseResourceProvider<R extends Resource> implements ResourceProvider<R> {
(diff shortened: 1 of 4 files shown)


fiat-web/src/main/java/com/netflix/spinnaker/fiat/config/DefaultResourcePermissionConfig.java
```diff
@@ -34,7 +35,7 @@
   @ConditionalOnProperty(
       value = "auth.permissions.source.account.resource.enabled",
       matchIfMissing = true)
-  @Order
+  @Order(Ordered.HIGHEST_PRECEDENCE + 100)
```
cfieber: at LOWEST_PRECEDENCE (the default) these are the last PermissionSources, which means that other sources (like ChaosMonkey) don't see the initial permission state from the resource configuration so they were always seeing an empty permission and doing nothing

fiat-roles/src/main/java/com/netflix/spinnaker/fiat/providers/BaseResourceProvider.java
```diff
@@ -86,6 +87,12 @@ public void setProviderCacheConfig(ProviderCacheConfig config) {
         .build();
   }
 
+  @Scheduled(fixedRateString = "PT#{${fiat.cache.expires-after-write-seconds:20} / 2}S")
+  public void reloadCache() {
```
cfieber: this just keeps the cache populated - there were occurrences where a user request to login would hit an expired cache and cause the reload to happen in the call flow, causing nasty outliers in response times

PR #630, opened 2020-04-03
chore(java11): Compile with Java 11 (but targeting Java 8)

plumpy:
Additional differences:

`UserRoleSyncer` was using `stream.map().count()` to run an operation on each element and then return the number of elements in the stream. But this doesn't work in Java 11 because "An implementation may choose to not execute the stream pipeline (either sequentially or in parallel) if it is capable of computing the count directly from the stream source." The Java 11 docs (unlike the Java 8 docs) [include an explicit example of this exact thing we're doing and why it doesn't work.](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/stream/Stream.html#count())

Fortunately we had to test to catch this. I wonder where else we're doing this... :/

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -13,9 +13,14 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
+      # Install Java 8 for cross-compilation support. Setting it up before
+      # Java 11 means it comes later in $PATH (because of how setup-java works)
       - uses: actions/setup-java@v1
         with:
           java-version: 8
+      - uses: actions/setup-java@v1
+        with:
+          java-version: 11
       - uses: actions/cache@v1
         with:
           path: ~/.gradle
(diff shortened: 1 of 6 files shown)


battlecow:
@spinnakerbot cherry-pick 1.19

ezimanyi:
@battlecow : There's an RFC with the timeline for our Java 11 migration [here](https://github.com/spinnaker/governance/blob/master/rfc/java11.md); this isn't something that we'll be cherry-picking back into older releases.

danopia:
@ezimanyi: No interest in at least cherry picking the actual breakage that was fixed here? https://github.com/spinnaker/fiat/pull/630/files#diff-f1ebc7cccfed4d99cad7f9887ed53912

Creating/updating service accounts (and thus using the pipeline permissions feature) is broken in 1.19 unless the operator switches to java8 images, which I don't think should be encouraged..

https://github.com/spinnaker/spinnaker/issues/5687
https://github.com/spinnaker/spinnaker/issues/5466#issuecomment-613080990

plumpy:
Oh, I can manually cherrypick that part. Thanks for pointing it out.

PR #632, opened 2020-04-06
fix(applications): fix dedupe of front50 and clouddriver apps

cfieber:
We were trying to dedupe on appname but each system presents them in a different case (front50 upper
clouddriver lower)


cfieber:
@spinnakerbot cherry-pick 1.19

PR #646, opened 2020-04-14
chore(build): update to newer spinnaker project plugin

cfieber:
this is a work in progress that applies spinnaker/spinnaker-gradle-project#121

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -6,6 +6,9 @@ on:
     - master
     - version-*
 
+env:
+  GRADLE_OPTS: -Dorg.gradle.daemon=false -Xmx2g -Xms2g
+
 jobs:
   branch-build:
     # Only run this on repositories in the 'spinnaker' org, not on forks.
@@ -28,8 +31,4 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-gradle-
       - name: Build
-        env:
-          BINTRAY_USER: ${{ secrets.BINTRAY_USER }}
-          BINTRAY_API_KEY: ${{ secrets.BINTRAY_API_KEY }}
-          GRADLE_OPTS: -Xmx6g -Xms6g -Dorg.gradle.daemon=false
-        run: ./gradlew -PenableCrossCompilerPlugin=true -PenablePublishing=true -PbintrayUser="${BINTRAY_USER}" -PbintrayKey="${BINTRAY_API_KEY}" build snapshot --stacktrace
+        run: ./gradlew -PenableCrossCompilerPlugin=true build --stacktrace
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -2,6 +2,9 @@ name: PR Build
 
 on: [ pull_request ]
 
+env:
+  GRADLE_OPTS: -Dorg.gradle.daemon=false -Xmx2g -Xms2g
+
 jobs:
   build:
     runs-on: ubuntu-latest
@@ -22,6 +25,4 @@ jobs:
         restore-keys: |
           ${{ runner.os }}-gradle-
     - name: Build
-      env:
-        GRADLE_OPTS: -Xmx6g -Xms6g -Dorg.gradle.daemon=false
       run: ./gradlew -PenableCrossCompilerPlugin=true build
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -6,6 +6,9 @@ on:
     - "v[0-9]+.[0-9]+.[0-9]+"
     - "v[0-9]+.[0-9]+.[0-9]+-rc.[0-9]+"
 
+env:
+  GRADLE_OPTS: -Dorg.gradle.daemon=false -Xmx2g -Xms2g
+
 jobs:
   release:
     runs-on: ubuntu-latest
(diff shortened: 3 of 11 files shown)


build.gradle
```diff
@@ -54,6 +50,11 @@ allprojects {
     apply plugin: 'java-library'
     apply plugin: 'groovy'
 
+    bintraySpinnaker {
+      enabled = true
+      bintrayOrg = 'cfieber'
```
cfieber, line 55: not strictly necessary - this extension will have reasonable defaults for our standard services

plumpy:
I'd be curious to see the changes to the GHA files, since we'll have to make equivalent changes [in the buildtool](https://github.com/spinnaker/buildtool/blob/master/dev/buildtool/cloudbuild/debs.yml). Is it as simple as adding a `-Pversion` flag?

cfieber:
it should be, but thanks for the pointer to the buildtool, I will make sure I haven't changed any of the bintray flag in use in there !

.github/workflows/release.yml
```diff
@@ -65,9 +58,9 @@ jobs:
            prerelease: ${{ steps.release_info.outputs.IS_CANDIDATE }}
       - name: Pause before dependency bump
         run: sleep 90
-      - name: Trigger dependency bump workflow
-        uses: peter-evans/repository-dispatch@v1
-        with:
-          token: ${{ secrets.SPINNAKER_GITHUB_TOKEN }}
-          event-type: bump-dependencies
-          client-payload: '{"ref": "${{ github.ref }}"}'
+#      - name: Trigger dependency bump workflow
```
cfieber, line 61: commented out while I was debugging things in my fork, will reenable

.github/workflows/pr.yml
```diff
@@ -22,6 +22,4 @@ jobs:
         restore-keys: |
           ${{ runner.os }}-gradle-
     - name: Build
-      env:
-        GRADLE_OPTS: -Xmx6g -Xms6g -Dorg.gradle.daemon=false
```
cfieber, line 26: I don't think this heap is necessary, gradle emits a warning message if it runs out of cache space and that is not showing up in any builds aside from the release build

.github/workflows/release.yml
```diff
@@ -31,26 +31,19 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
-          . .github/workflows/release_info.sh ${{ github.event.repository.name }}
+          . .github/workflows/release_info.sh ${{ github.event.repository.full_name }}
```
cfieber, line 34: was testing in my fork, this provides `cfieber/fiat` instead of just `fiat`

.github/workflows/release.yml
```diff
@@ -31,26 +31,19 @@ jobs:
...
           echo ::set-output name=CHANGELOG::$(echo -e "${CHANGELOG}")
           echo ::set-output name=SKIP_RELEASE::${SKIP_RELEASE}
           echo ::set-output name=IS_CANDIDATE::${IS_CANDIDATE}
-      - name: Candidate build
```
cfieber, line 38: We aren't doing anything meaningfully different in the release between a -rc tag and a regular tag

.github/workflows/pr.yml
cfieber: actually there is a magical blend of `-Xmx and org.gradle.daemon` here that makes the build not fork off a daemon

6g is definitely way more than we need but this setting is effective/useful so I'll put it back in with a smaller heap

.github/workflows/build.yml
```diff
@@ -28,8 +28,4 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-gradle-
       - name: Build
-        env:
-          BINTRAY_USER: ${{ secrets.BINTRAY_USER }}
-          BINTRAY_API_KEY: ${{ secrets.BINTRAY_API_KEY }}
-          GRADLE_OPTS: -Xmx6g -Xms6g -Dorg.gradle.daemon=false
-        run: ./gradlew -PenableCrossCompilerPlugin=true -PenablePublishing=true -PbintrayUser="${BINTRAY_USER}" -PbintrayKey="${BINTRAY_API_KEY}" build snapshot --stacktrace
```
cfieber, line 35: We previously build and published -SNAPSHOT versions which would go to the oss.jfrog.org snapshot repo.

I'm not aware of any usage of the snapshot versions, so I'd prefer not to add support for this

.github/workflows/pr.yml
cfieber: I've moved this to a top level env block in all the actions

.github/workflows/release_info.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash -x
 
 # Only look to the latest release to determine the previous tag -- this allows us to skip unsupported tag formats (like `version-1.0.0`)
-export PREVIOUS_TAG=`curl --silent "https://api.github.com/repos/spinnaker/$1/releases/latest" | grep '"tag_name":' | sed -E 's/.*"([^"]+)".*/\1/'`
```
cfieber, line 4: passing full_name instead of name now

.github/workflows/release.yml
jonsie: Ah nice change.

.github/workflows/release.yml
jonsie: ✂️ 🔪 💯

PR #663, opened 2020-04-27
feat(plugins): add plugins to fiat

danielpeach:
This PR makes Fiat pluggable, similar to spinnaker/clouddriver#4383

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ build/
 out/
 /.idea/
 /.shelf/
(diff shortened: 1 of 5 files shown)


danielpeach:
We wanted to do some noodling on adding custom `ResourceProviders` before adding a real extension point there.

Not actually super clear to me what Fiat client should do in terms of plugins. Should it run pf4j separately and fetch its own fiat plugins?

PR #668, opened 2020-05-04
feat(serviceAccounts): enable customization of service account resource provider

cfieber:
introduces BaseServiceAccountResourceProvider to capture service account specific
access checks"


fiat-web/src/main/java/com/netflix/spinnaker/fiat/config/FiatConfig.java
```diff
@@ -88,6 +90,15 @@ DefaultApplicationResourceProvider applicationProvider(
         properties.isAllowAccessToUnknownApplications());
   }
 
+  @Bean
+  @ConditionalOnProperty(
+      value = "fiat.service-account-resource-provider.default.enabled",
```
cfieber, line 95: I tried a few things to make this just `@ConditionalOnBean` but we would have to rework a bunch of our things away from initializing via `@ComponentScan` of `com.netflix.spinnaker.config. If everything was loaded via autoconfigurations I think it could work, but I don't really want to boil that ocean at the moment.

jonsie: Yeah I've tried this before and thrown my hands up.

PR #679, opened 2020-05-08
config(core): Move redis connection default to base config

ezimanyi:
Currently Halyard is appending a default for the redis config to all configs it generates. Let's push this down to the base config so all users get the same defaults and we don't need to append this all the time.

Currently the base config uses 'services.redis.connection' as the default. This doesn't match what is used for other services, nor what is used for the redis connection in other microservices (which use services.redis.baseUrl).

This commit changes this to look at 'services.redis.baseUrl' for consistency. Users who have configured a 'services.redis.connection' property but not 'services.redis.baseUrl' (and who are depending on that rather than just setting redis.connection directly in their gate config) will need to update to account for this change.

### DIFF
--- a/fiat-web/config/fiat.yml
+++ b/fiat-web/config/fiat.yml
@@ -2,7 +2,7 @@ server:
   port: 7003
 
 redis:
-  connection: ${services.redis.connection:redis://localhost:6379}
+  connection: ${services.redis.baseUrl:redis://localhost:6379}
 
 okHttpClient:
   retries:
--- a/halconfig/README.md
+++ b/halconfig/README.md
@@ -1,0 +1,6 @@
+This directory contains skeleton fiat configs to which Halyard concatenates
+its generated deployment-specific config.
+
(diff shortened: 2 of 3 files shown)


ezimanyi:
Would like someone from Netflix to weigh in on the change from looking at `services.redis.connection` to `services.redis.baseUrl`.  Everywhere else (including the redis connections in other services) we're looking at `services.redis.baseUrl` so I changed this for consistency...but definitely want to check if you are depending on this at all.

If we want to leave this we could do a more complex expression (`${${services.redis.connection:services.redis.baseUrl}:redis://localhost:6379}`) but I'd rather avoid that if this is just a relic of some old config that we can get rid of.

ezimanyi:
@robzienert @cfieber @jonsie : Any thoughts on this?

cfieber:
I'm all for consistency, we can sort out our configs as needed if this lands

cfieber:
(we don't use any of the yml configuration in the project, we have our own config file for fiat at set redis.connection directly in there)

PR #682, opened 2020-05-13
fix(misc): Use new client provider and remove usage of deprecated OkH…

srekapalli:
- Update to use OkHttp3 and the new provider implementation.


plumpy:
> There is a chance the autobump PR to gate will fail and it will need a bit of manual intervention.

Looks like 3 succeeded and 5 failed. Here's a quick link to the failures:
https://github.com/pulls?q=is%3Aopen+is%3Apr+author%3Aspinnakerbot+fiatVersion

PR #683, opened 2020-05-13
fix(serviceAccounts): limit the number of service accounts to expand

cfieber:
This is primarily a guard against a user with the admin flag (who has access to all service accounts) in a Spinnaker instance with hundreds of service accounts. This endpoint can quickly get up past the point where the request from gate will time out (and fall back to just asking for the unexpanded list).


cfieber:
.

fiat-web/src/main/java/com/netflix/spinnaker/fiat/controllers/AuthorizeController.java
```diff
@@ -140,6 +141,13 @@ public AuthorizeController(
       return serviceAccounts;
     }
 
+    if (serviceAccounts.size() > configProps.getMaxExpandedServiceAccounts()) {
+      throw new InvalidRequestException(
+          String.format(
+              "Unable to expand service accounts for user %s. User has %s service accounts. Maximum expandable service accounts is %s.",
```
jonsie, line 147: 👍

PR #688, opened 2020-05-19
chore(dynomite): Remove dynomite from project

robzienert:


fiat-roles/src/main/java/com/netflix/spinnaker/fiat/config/RedisConfig.java
```diff
@@ -6,13 +6,13 @@
 import com.netflix.spinnaker.kork.jedis.telemetry.InstrumentedJedisPool;
 import java.net.URI;
 import lombok.extern.slf4j.Slf4j;
-import org.apache.commons.lang3.StringUtils;
```
robzienert, line 9: Without dynomite, commons-lang3 is unavailable, so I replaced with Spring.

fiat-roles/src/main/java/com/netflix/spinnaker/fiat/config/RedisConfig.java
```diff
@@ -52,7 +52,7 @@ private static JedisPool createPool(
 
     String path = redisConnection.getPath();
     if (StringUtils.isEmpty(path)) {
-      path = "/" + String.valueOf(Protocol.DEFAULT_DATABASE);
+      path = "/" + Protocol.DEFAULT_DATABASE;
```
robzienert, line 55: IntelliJ said to cleanup, so I cleaned up.

PR #690, opened 2020-05-20
fix(misc): Move to client provider impl for spinnaker internal services

srekapalli:
Move to client provider impl for spinnaker internal services


ajordens:
LGTM but I'll defer final approval to someone else on the list.

I did spot check that the `retryOnConnectionFailure` lives on through the `OkHttpClient` created by `RawOkHttpClientConfiguration` right?

srekapalli:
> LGTM but I'll defer final approval to someone else on the list.
> 
> I did spot check that the `retryOnConnectionFailure` lives on through the `OkHttpClient` created by `RawOkHttpClientConfiguration` right?

Yes. It is on the RawOkHttpClient and we use configuration prop to set that value.

PR #710, opened 2020-06-17
chore(dependencies): Autobump korkVersion

spinnakerbot:


jonsie:
@cfieber Looks like some additional fiat cleanup is required -- there are references here to `sslExtensionConfigurationProperties`.

PR #712, opened 2020-06-18
Extension Resources

danielpeach:
This change allows Fiat to serve / evaluate permissions for arbitrary resources. I've added an example plugin here that demonstrates how it works: https://github.com/spinnaker-plugin-examples/fileResourceProvider. That example is Spring-based. Ideally I'd get this PR merged and then pull out `Resource` and `ResourceProvider` extension points.

The resulting map served from `/authorize/{user}` looks like this:

```
{
  ...
  "extensionResources": {
     "my_extension_resource": [{
         "name": ...
         "authorizations": ["READ", "WRITE"]
     }]
  },
  ...
}
```

There's a slight tweak in fiat client code to automatically evaluate authorizations on these resources.

### DIFF
--- a/fiat-core/src/main/java/com/netflix/spinnaker/fiat/model/UserPermission.java
+++ b/fiat-core/src/main/java/com/netflix/spinnaker/fiat/model/UserPermission.java
@@ -17,23 +17,14 @@
 package com.netflix.spinnaker.fiat.model;
 
 import com.fasterxml.jackson.annotation.JsonIgnore;
-import com.netflix.spinnaker.fiat.model.resources.Account;
-import com.netflix.spinnaker.fiat.model.resources.Application;
-import com.netflix.spinnaker.fiat.model.resources.BuildService;
-import com.netflix.spinnaker.fiat.model.resources.Resource;
-import com.netflix.spinnaker.fiat.model.resources.Role;
-import com.netflix.spinnaker.fiat.model.resources.ServiceAccount;
-import com.netflix.spinnaker.fiat.model.resources.Viewable;
-import java.util.Collection;
-import java.util.Collections;
-import java.util.HashSet;
-import java.util.LinkedHashSet;
-import java.util.Set;
+import com.netflix.spinnaker.fiat.model.resources.*;
+import java.util.*;
 import java.util.function.Function;
 import java.util.stream.Collectors;
 import lombok.Data;
 import lombok.EqualsAndHashCode;
 import lombok.NoArgsConstructor;
+import lombok.val;
 
 @Data
 public class UserPermission {
@@ -44,6 +35,7 @@ public class UserPermission {
   private Set<ServiceAccount> serviceAccounts = new LinkedHashSet<>();
   private Set<Role> roles = new LinkedHashSet<>();
   private Set<BuildService> buildServices = new LinkedHashSet<>();
+  private Set<Resource> extensionResources = new LinkedHashSet<>();
   private boolean admin = false;
 
   public void addResource(Resource resource) {
@@ -68,7 +60,7 @@ public class UserPermission {
           } else if (resource instanceof BuildService) {
             buildServices.add((BuildService) resource);
           } else {
-            throw new IllegalArgumentException("Cannot add unknown resource " + resource);
+            extensionResources.add(resource);
           }
         });
 
@@ -83,6 +75,7 @@ public class UserPermission {
     retVal.addAll(serviceAccounts);
     retVal.addAll(roles);
     retVal.addAll(buildServices);
+    retVal.addAll(extensionResources);
     return retVal;
   }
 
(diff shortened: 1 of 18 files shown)


fiat-core/src/main/java/com/netflix/spinnaker/fiat/model/UserPermission.java
```diff
@@ -131,6 +125,18 @@ public View(UserPermission permission) {
...
       this.roles = (Set<Role.View>) toViews.apply(permission.getRoles());
       this.buildServices = (Set<BuildService.View>) toViews.apply(permission.getBuildServices());
+
+      this.extensionResources = new HashMap<>();
+      for (val resource : permission.extensionResources) {
+        val set =
+            this.extensionResources.computeIfAbsent(
+                resource.getResourceType(), (ignore) -> new HashSet<>());
+        val view = ((Viewable) resource).getView(permission.getRoles(), permission.isAdmin());
+        if (view instanceof Authorizable) {
```
danielpeach, line 135: This is not great. The types need to be re-worked at some point. When you push a `Viewable` resource into `UserPermission`, you have no idea what kind of view it's going to give you! You have to do this check here to make sure that it was actually an `Authorizable`.

Ideally the plugin developer would know at compile-time what type their resource has to implement. Right now, they just have to magically know that their `Resource`'s `View` has to implement `Authorizable`.

fiat-core/src/main/java/com/netflix/spinnaker/fiat/model/resources/ResourceType.java
```diff
@@ -16,32 +16,60 @@
...
+  public static ResourceType BUILD_SERVICE = new ResourceType("build_service");
+
+  public ResourceType(String name) {
+    this.name = name;
   }
 
   // TODO(ttomsu): This is Redis-specific, so it probably shouldn't go here.
-  public static ResourceType parse(@Nonnull String pluralOrKey) {
+  public static ResourceType parse(String pluralOrKey) {
+    if (pluralOrKey == null) {
```
danielpeach, line 38: This is explicitly tested for. Not sure why, but didn't want to change anything.

fiat-core/src/main/java/com/netflix/spinnaker/fiat/model/resources/ResourceType.java
```diff
@@ -16,32 +16,60 @@
...
+    }
     pluralOrKey = pluralOrKey.substring(pluralOrKey.lastIndexOf(':') + 1);
     String singular =
         pluralOrKey.endsWith("s")
             ? pluralOrKey.substring(0, pluralOrKey.length() - 1)
             : pluralOrKey;
-    return ResourceType.valueOf(singular.toUpperCase());
+
+    if ("".equals(singular)) {
+      throw new IllegalArgumentException(
```
danielpeach, line 48: Same here.

fiat-core/src/main/java/com/netflix/spinnaker/fiat/model/UserPermission.java
cfieber: can we update View to extend/implement Authorizable in Resource or is there no common hierarchy here?

fiat-core/src/main/java/com/netflix/spinnaker/fiat/model/resources/ResourceType.java
cfieber: hmm, I think previously or at some point the @Nonnull was lombok's annotation rather than javax.annotation that causes it to throw on a null argument

cfieber:
One thing to consider with fiat is that fiat-api is meant as the tool for other services to connect to fiat, not an api module for plugin developers

When you get to the point of starting to pull out a plugin developer's api module (if that was your next step) we'll need to be aware of the current dependency all the other (authz'ing at least) services have on fiat

fiat-core/src/main/java/com/netflix/spinnaker/fiat/model/resources/ResourceType.java
danielpeach: Yeah I wanted to switch to the lombok annotation, but the exception type for `@NonNull` is configured in `lombok.config` is `IllegalArgumentException`.

fiat-core/src/main/java/com/netflix/spinnaker/fiat/model/UserPermission.java
danielpeach: I wish! The problem is that `ServiceAccount.View` and `Role.View` do their own thing and don't implement `Authorizable`.

danielpeach:
@cfieber I have a branch [here](https://github.com/spinnaker/fiat/compare/master...danielpeach:dp/hardcode-responses?expand=1) that re-writes `AuthorizeControllerSpec` to check against hardcoded responses from `/authorize/{user}`. The tests pass against Fiat@head and pass without any changes if I cherry-pick those changes into this PR. 

I've also manually tested with Clouddriver@head against these changes.

PR #717, opened 2020-06-23
chore(build): gradle 6.5

cfieber:


settings.gradle
```diff
@@ -23,6 +23,8 @@ if (spinnakerGradleVersion.endsWith('-SNAPSHOT')) {
   }
 }
 
+enableFeaturePreview("VERSION_ORDERING_V2")
+
```
srekapalli, line 27: Do we need to enable this feature ?

cfieber: Since this will become the default I'd like to catch any wierdness that may arise out of this early.

That said I haven't seen any issues as a result, I think it will only matter for resolving -SNAPSHOT and such

PR #722, opened 2020-06-25
Add Fiat ExtensionPoints

danielpeach:
This is a follow-up on https://github.com/spinnaker/fiat/pull/712. Adds two extension points: `Resource` and `ResourceLoader`.

`ResourceLoader` is a simplification of the `ResourceProvider` interface. The `ResourceLoaders` get pulled into a new `ResourceProviderRegistry`, which maps the loaders into `ResourceProvider`s.

`Resource` and `ResourceLoader` are now in `fiat-api` along with a few of the interfaces they depend on. The old `fiat-api` is now `fiat-client`. Once this is ready to be merged I'll open PRs to update the services that depend on `fiat-client`.


fiat-api/src/main/java/com/netflix/spinnaker/fiat/api/Permissions.java
```diff
@@ -79,13 +77,8 @@ public boolean isRestricted() {
     return this.permissions.values().stream().anyMatch(groups -> !groups.isEmpty());
   }
 
-  public boolean isAuthorized(Set<Role> userRoles) {
-    return !getAuthorizations(userRoles).isEmpty();
-  }
-
-  public Set<Authorization> getAuthorizations(Set<Role> userRoles) {
```
danielpeach, line 86: `Permission` used to depend on `Role` which in turn brought in a bunch of stuff that I didn't want to pull into `fiat-api`.

fiat-api/src/main/java/com/netflix/spinnaker/fiat/api/Resource.java
```diff
@@ -14,17 +14,25 @@
...
+package com.netflix.spinnaker.fiat.api;
 
 import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.netflix.spinnaker.kork.annotations.Alpha;
+import org.pf4j.ExtensionPoint;
 
-public interface Resource {
+/** Represents a Fiat resource (e.g., a Role or a Service account). */
+@Alpha
+public interface Resource extends ExtensionPoint {
```
danielpeach, line 25: I had problems deserializing these if they extended `SpinnakerExtensionPoint` and didn't want to make this PR any larger by updating the object mapper. I'll come back to this.

fiat-core/src/main/java/com/netflix/spinnaker/fiat/model/UserPermission.java
```diff
@@ -131,10 +136,16 @@ public View(UserPermission permission) {
         val set =
             this.extensionResources.computeIfAbsent(
                 resource.getResourceType(), (ignore) -> new HashSet<>());
-        val view = ((Viewable) resource).getView(permission.getRoles(), permission.isAdmin());
-        if (view instanceof Authorizable) {
-          set.add((Authorizable) view);
+        val builder = AuthorizableImpl.builder();
```
danielpeach, line 139: All of the built-in `Resource`s follow a `Viewable` -> `Resource.View` pattern which I didn't think a plugin developer should have to deal with. This code just maps a `Resource.AccessControlled` into an `Authorizable`, which is all the `Viewable` stuff was ever doing.

fiat-plugins-test/fiat-plugins-test.gradle
```diff
@@ -0,0 +1,71 @@
...
+  testRuntimeOnly("org.junit.platform:junit-platform-launcher")
+  testRuntimeOnly("org.junit.jupiter:junit-jupiter-engine")
+}
+
+test {
+  useJUnitPlatform {
+    includeEngines("junit-jupiter", "junit-vintage")
+  }
+  filter {
+    includeTestsMatching "com.netflix.spinnaker.fiat.plugins.FiatPluginsTest"
```
danielpeach, line 62: This is the same stuff that @jonsie ran into in this PR: https://github.com/spinnaker/orca/pull/3579

fiat-plugins-test/fiat-plugins-test.gradle
```diff
@@ -0,0 +1,71 @@
+/*
```
danielpeach, line 1: Open to breaking out the plugin TCK into a separate PR but wanted to be sure everything worked all the way through.

fiat-web/fiat-web.gradle
```diff
@@ -48,4 +49,21 @@ dependencies {
   testImplementation "com.netflix.spinnaker.kork:kork-jedis-test"
 }
 
+test {
```
danielpeach, line 52: Same as here: https://github.com/spinnaker/orca/blob/master/orca-web/orca-web.gradle#L104

Not great but needed for the plugins TCK to work

fiat-web/src/main/java/com/netflix/spinnaker/fiat/Main.java
```diff
@@ -44,6 +45,7 @@
       // https://github.com/spring-projects/spring-ldap/issues/473
       LdapHealthContributorAutoConfiguration.class
     })
+@SpringBootApplication
```
danielpeach, line 48: Redundant but also needed for the tck. `SpringPackageVersionResolver` doesn't work without it: https://github.com/spinnaker/kork/blob/master/kork-core/src/main/java/com/netflix/spinnaker/kork/version/SpringPackageVersionResolver.java#L44

fiat-api/src/main/java/com/netflix/spinnaker/fiat/api/Permissions.java
robzienert: Looking at `Role`, it's not immediately obvious to me what extra things would be brought alongside?

fiat-api/src/main/java/com/netflix/spinnaker/fiat/api/Resource.java
robzienert: The consequences of not using `SpinnakerExtensionPoint` is that the proxy will not work (metrics, logging, etc). I'd be interested to hear more about the serialization issues you're seeing. I suspect it's the fact that `getExtensionClass()` is automatically picked up as a property for Jackson?

I think the right resolution (if my assumption is correct) would be to change `SpinnakerExtensionPoint` to rename `getExtensionClass()` to something that won't be automatically picked up by Jackson.

fiat-core/src/main/java/com/netflix/spinnaker/fiat/model/UserPermission.java
robzienert: Agree, nice call.

fiat-plugins-test/fiat-plugins-test.gradle
robzienert: No - this makes the PR complete imo. Nice.

cfieber:
what does rollout for this change look like? This will be compile time breaking for all the services, but do any of these refactors change fiat's API payloads? (Role -> String maybe comes to mind)

I'm not too worried if we need to redeploy things in a particular order (first do fiat with these changes, then services, or the reverse) but we do need to be able to roll this in without any downtime

PR #745, opened 2020-08-12
chore(dockerfile): various Dockerfile cleanups

plumpy:


Dockerfile.slim
```diff
@@ -2,7 +2,8 @@ FROM alpine:3.11
 LABEL maintainer="sig-platform@spinnaker.io"
 COPY fiat-web/build/install/fiat /opt/fiat
 RUN apk --no-cache add --update bash openjdk11-jre
-RUN adduser -D -S spinnaker
+RUN addgroup -S -g 10111 spinnaker
+RUN adduser -S -G spinnaker -u 10111 spinnaker
```
maggieneterval, line 5: Do we need `spinnaker` to be a system user (also thought there was some max to system user IDs that fell below 10111)?

plumpy: A system user just means you can't login as that user (it has no password and no shell), so it seems more secure. By default they're created with low UIDs but there's no particular reason it needs to.

maggieneterval: Oh okay, sounds good, thanks! 👍 🎉

PR #747, opened 2020-08-18
perf(auth/ldap): Limit search object size

sikevux:
Currently when fiat sync LDAP roles with the new Group -> User mapping
the response is huge, as mentioned in the original commit. To workaround
this issue this change adds the wanted attributes to the search, making
the search only return those attributes.

To be able to use "attributes to return" feature we also have to set a
search scope. Using the default scope of SUBTREE_SCOPE caused errors on
my test setup, and changing to OBJECT_SCOPE solved those issues. Unit tests
however still pass in both cases.

The delimiter for the GroupUserAttributes is configurable, but will
default to space, as it's the default for ldapsearch.


sikevux:
@spinnaker/reviewers

cfieber (commented):
seems reasonable to me but I don't use LDAP, would love a thumbs-up from someone who's using it in production

cfieber:
@spinnaker/oss-reviewers PTAL if you are using LDAP

@sikevux I'll merge this if no one comments otherwise, sorry for the delay on it

cfieber (approved):

PR #755, opened 2020-08-25
fix(google): Retrieve all the groups for a user

jfrabaute:
The list of groups is paginated (default page is 200), so when a user
has more than 200 groups, only the first 200 are retrieve.
This change is retrieving all the groups for a user by querying all the
pages.


jfrabaute:
I wanted to test the change.
I've started a dev env on my laptop, following the guide with LocalGit and "hal", but front50 does not want to start (I've setup "redis" as persistent storage).
I've posted on #dev spinnaker Slack group.
Hopefully I get an answer on how to start everything on a laptop to test.

fiat-google-groups/src/main/java/com/netflix/spinnaker/fiat/roles/google/GoogleDirectoryUserRolesProvider.java
```diff
@@ -174,8 +174,23 @@ public void onSuccess(Groups groups, HttpHeaders responseHeaders) throws IOExcep
...
+    while (nextPageToken != null) {
+      final Groups nextPage =
+          service
+              .groups()
+              .list()
+              .setDomain(config.getDomain())
+              .setUserKey(email)
+              .setPageToken(nextPageToken)
+              .execute();
+      groups.getGroups().addAll(nextPage.getGroups());
```
cfieber, line 190: I don't know that you could rely on the List<Group> from getGroups being mutable. Might be worth keeping a separate List<Group> that you accumulate into and then call groups.setGroups with that at the end just in case

jfrabaute: Good point.
I'm going to work on this and update the PR if needed.

Thanks for the review btw.

jfrabaute: I tested and it works fine, so the list from getGroups is mutable.

jfrabaute: @cfieber Are you ok with the current code? Or do you want to change it?

cfieber (approved):

PR #758, opened 2020-08-26
feat(metrics): instrument PermissionRepository

cfieber:
Adds a Configuration class to create RedisPermissionRepository and uses InstrumentedProxy to add
method invocation metrics

### DIFF
--- a/fiat-roles/src/main/java/com/netflix/spinnaker/fiat/config/PermissionsRepositoryConfig.java
+++ b/fiat-roles/src/main/java/com/netflix/spinnaker/fiat/config/PermissionsRepositoryConfig.java
@@ -1,0 +1,33 @@
+package com.netflix.spinnaker.fiat.config;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.netflix.spectator.api.Registry;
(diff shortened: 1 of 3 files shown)


cfieber:
this was part of something else I have in progress but I'd like to push it out to get before/after metrics

fiat-roles/src/main/java/com/netflix/spinnaker/fiat/config/PermissionsRepositoryConfig.java
```diff
@@ -0,0 +1,33 @@
...
+      RedisClientDelegate redisClientDelegate,
+      List<Resource> resources,
+      @Value("${fiat.redis.prefix:spinnaker:fiat}") String prefix) {
+    PermissionsRepository repository =
+        new RedisPermissionsRepository(objectMapper, redisClientDelegate, resources, prefix);
+    return InstrumentedProxy.proxy(
+        registry,
+        repository,
+        "permissionsRepository",
+        Collections.singletonMap("repositoryType", "redis"));
```
robzienert, line 31: yee nice.

PR #764, opened 2020-09-02
fix(repo): don't swallow exceptions in PermissionRepository.get

cfieber:
Allows RedisPermissionRepository.get to fail with an exception
instead of masking failure with an Optional.empty

This will allow better fallback behaviour and retries instead
of just appearing as an unknown user.

Adds retries for reading user data from redis to handle pool
exhaustion or redis timeouts and introduces an overall
timeout for the user read in case to prevent excessive retries
from introducing unacceptable latency.

The RedisPermissionRepository.get by default is set to
time out after 1 second but this value is configurable with
the property 'fiat.redis.repository.get-permission-timeout'

There is also a timeout for the call to determine whether the
UNRESTRICTED user has changed. By default this is 50 milliseconds
but is configurable with the property
'fiat.redis.repository.check-last-modified-timeout'


fiat-roles/src/main/java/com/netflix/spinnaker/fiat/permissions/RedisPermissionsRepository.java
```diff
@@ -76,43 +82,120 @@
...
+      return clock.millis() > timeoutAtMs;
+    }
+
+    String getName() {
+      return name;
+    }
+  }
+
+  private <T> T redisRead(TimeoutContext timeoutContext, Function<JedisCommands, T> fn) {
+    return retryRegistry
```
cfieber, line 174: need to do the retry this way as opposed to annotation based because the AOP stuff doesn't kick in on an internal method call within a class

fiat-roles/src/main/java/com/netflix/spinnaker/fiat/permissions/RedisPermissionsRepository.java
```diff
@@ -76,43 +82,120 @@
...
+        .retry("permissionsRepositoryRedisRead")
```
cfieber, line 175: derp I added a constant for this name and forgot to update here will fix

fiat-roles/src/main/java/com/netflix/spinnaker/fiat/permissions/RedisPermissionsRepository.java
```diff
@@ -76,43 +82,120 @@
...
+            () -> {
+              if (timeoutContext.isTimeout()) {
+                throw new PermissionReadException(
+                    String.format("request processing timeout for %s", timeoutContext.getName()));
+              }
+              return redisClientDelegate.withCommandsClient(fn);
+            });
+  }
+
+  private Map<String, String> hgetall(TimeoutContext timeoutContext, String key) {
```
cfieber, line 186: hgetall functionality via hscan instead - if we get a humongous hash for apps our accounts this will be more but smaller/faster redis interactions

fiat-roles/src/main/java/com/netflix/spinnaker/fiat/permissions/RedisPermissionsRepository.java
```diff
@@ -76,43 +82,120 @@
...
+                clock,
+                configProps.getRepository().getCheckLastModifiedTimeout()),
+            c -> c.get(unrestrictedLastModifiedKey()));
     if (serverLastModified == null) {
       serverLastModified = NO_LAST_MODIFIED;
     }
     return unrestrictedPermission.get(serverLastModified);
   }
 
+  private static class TimeoutContext {
```
cfieber, line 149: I'll add a doc comment on this class, but essentially it is here to bail on trying to load a user if too much elapsed time has passed. Because this is doing retries at the individual redis read level instead of a retry around the whole get method in the case of connection pool contention or other redis slowness I don't want to eventually succeed if the result is useless at that point - it would be better to bail out with an error and let external retries or fallbacks kick in

fiat-roles/src/main/java/com/netflix/spinnaker/fiat/permissions/RedisPermissionsRepository.java
```diff
@@ -206,40 +288,36 @@ public RedisPermissionsRepository put(@NonNull UserPermission permission) {
...
-                    r -> {
-                      String userKey = userKey(id, r);
-                      Response<Map<String, String>> resourceMap = p.hgetAll(userKey);
-                      userResponseMap.put(r, resourceMap);
-                    });
-            Response<Boolean> admin = p.sismember(adminKey(), id);
-            p.sync();
-            userResponseMap.isAdmin = admin.get();
+      UserPermission userPermission = new UserPermission().setId(id);
+      resources.forEach(
```
cfieber, line 302: I don't think the pipeline to fetch all the things and then deserialize later would be meaningfully faster than just executing the appropriate calls, and this has the added benefit of only retrying some portion of the call that failed due to a jedis exception instead of re-invoking the whole pipeline

fiat-roles/src/main/java/com/netflix/spinnaker/fiat/permissions/RedisPermissionsRepository.java
```diff
@@ -206,40 +288,36 @@ public RedisPermissionsRepository put(@NonNull UserPermission permission) {
...
       return Optional.of(userPermission);
+    } catch (SpinnakerException se) {
+      throw se;
     } catch (Exception e) {
-      log.error("Storage exception reading " + id + " entry.", e);
+      String message = String.format("Storage exception reading %s entry.", id);
+      log.error(message, e);
+      throw new PermissionReadException(message, e);
     }
-    return Optional.empty();
```
cfieber, line 242: no longer masking exceptions and treating them as a user not found

I'm reasonably happy with introducing more error responses from the API in this change because I think it is more correct / less surprising than just failing with a not found and there are retries on all the read operations so the case where the not-found would kick into our fallback in authorize controller as a fallback for a redis exception (rather than just as a fallback for the user not having transited gate) I think is better handled in here with retrying the read operations against redis

fiat-roles/src/test/groovy/com/netflix/spinnaker/fiat/permissions/RedisPermissionsRepositorySpec.groovy
```diff
@@ -29,53 +29,151 @@ import com.netflix.spinnaker.fiat.model.resources.Role
...
+        configProps,
+        RetryRegistry.ofDefaults()
     )
   }
 
   def cleanup() {
     jedis.flushDB()
   }
 
+  def "should fail if timeout is exceeded"() {
```
cfieber, line 147: woof. sorry about this test I don't really like it!

robzienert: It's looks great.

fiat-web/src/test/groovy/com/netflix/spinnaker/fiat/controllers/AuthorizeControllerSpec.groovy
```diff
@@ -31,6 +31,7 @@ import com.netflix.spinnaker.fiat.model.resources.Resource
 import com.netflix.spinnaker.fiat.permissions.PermissionsRepository
 import com.netflix.spinnaker.fiat.permissions.PermissionsResolver
 import com.netflix.spinnaker.fiat.providers.ResourcePermissionProvider
+import com.netflix.spinnaker.kork.web.exceptions.NotFoundException
```
robzienert, line 34: Unused?

robzienert (approved):

fiat-roles/src/main/java/com/netflix/spinnaker/fiat/permissions/RedisPermissionsRepository.java
jonsie: Agreed.  This is way cleaner and better.  Nice improvement.

fiat-roles/src/main/java/com/netflix/spinnaker/fiat/permissions/RedisPermissionsRepository.java
```diff
@@ -487,4 +515,60 @@ default R apply(T t) {
...
+    boolean isTimeout() {
+      return clock.millis() > timeoutAtMs;
+    }
+
+    String getName() {
+      return name;
+    }
+  }
+
+  private <T> T redisRead(TimeoutContext timeoutContext, Function<JedisCommands, T> fn) {
```
jonsie, line 549: Cool!

fiat-roles/src/test/groovy/com/netflix/spinnaker/fiat/permissions/RedisPermissionsRepositorySpec.groovy
```diff
@@ -29,53 +29,151 @@ import com.netflix.spinnaker.fiat.model.resources.Role
...
+      latch.countDown()
+      repo.get("foo")
+    }
+
+    latch.await()
+
+    //TODO(cfieber): apologies to all readers of this test code
+    // this is gross, but a brief sleep will ensure the repo.get
+    // call attempts to make a redis call and hits the pause
+    // point in the PauseableDelegate.
```
jonsie, line 165: Not sure `PauseableDelegate` is a thing.

jonsie: Oh - `PausableRedisClientDelegate`.

fiat-roles/src/main/java/com/netflix/spinnaker/fiat/permissions/RedisPermissionsRepository.java
```diff
@@ -487,4 +515,60 @@ default R apply(T t) {
...
+
+  /**
+   * TimeoutContext allows specifying an expiration time for request processing.
+   *
+   * <p>If something exceeds the specified timeout duration, the request handler should stop doing
+   * work and just bail out with a timeout.
+   */
+  private static class TimeoutContext {
+    private final String name;
+    private final long timeoutAtMs;
```
jonsie, line 527: big time nit: `timeoutMs` seems more idiomatic than `timeoutAtMs`, not sure the `At` achieves much (though I know what you're going for).

fiat-roles/src/main/java/com/netflix/spinnaker/fiat/permissions/PermissionReadException.java
```diff
@@ -0,0 +1,18 @@
...
+
+/**
+ * An exception reading permission values.
+ *
+ * <p>Reading permissions is side-effect free so this exception is retryable by default.
+ */
+public class PermissionReadException extends PermissionRepositoryException {
+  public PermissionReadException(String message) {
+    super(message);
+    setRetryable(true);
```
jonsie, line 11: So is the idea that the `retryable` field in this exception would bubble up to the caller and we could retry somewhere else (like maybe a retry across the wire?).

fiat-roles/src/test/groovy/com/netflix/spinnaker/fiat/permissions/RedisPermissionsRepositorySpec.groovy
cfieber: oops, yeah I renamed and apparently it didn't find the usage in the comment, thanks!

fiat-web/src/test/groovy/com/netflix/spinnaker/fiat/controllers/AuthorizeControllerSpec.groovy
cfieber: there previously was a c.n.s.f.controllers.NotFoundException (same package as this test) that I deleted in favor of the one from kork-exceptions

fiat-roles/src/main/java/com/netflix/spinnaker/fiat/permissions/RedisPermissionsRepository.java
cfieber: I would tend to see a value timeoutMs as something more like a duration than an instant

cfieber: Actually going to just change this to an Instant, I suspect this is just stupid over optimization!

fiat-roles/src/main/java/com/netflix/spinnaker/fiat/permissions/PermissionReadException.java
jonsie: Cool, gotcha.

jonsie (approved):

PR #765, opened 2020-09-02
fix(test): add slightly longer sleep time to poorly concurrent test

cfieber:


cfieber:
saw this fail in a github actions build

PR #767, opened 2020-09-08
fix(perms): adds fallback for fetching unrestricted user

cfieber:
Also adds a bunch of logging around the /authorize flow

### DIFF
--- a/fiat-roles/src/main/java/com/netflix/spinnaker/fiat/permissions/RedisPermissionsRepository.java
+++ b/fiat-roles/src/main/java/com/netflix/spinnaker/fiat/permissions/RedisPermissionsRepository.java
@@ -30,6 +30,7 @@ import com.netflix.spinnaker.fiat.model.UserPermission;
 import com.netflix.spinnaker.fiat.model.resources.Resource;
 import com.netflix.spinnaker.fiat.model.resources.ResourceType;
 import com.netflix.spinnaker.fiat.model.resources.Role;
+import com.netflix.spinnaker.kork.exceptions.IntegrationException;
 import com.netflix.spinnaker.kork.exceptions.SpinnakerException;
 import com.netflix.spinnaker.kork.jedis.RedisClientDelegate;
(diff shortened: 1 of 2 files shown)


fiat-roles/src/main/java/com/netflix/spinnaker/fiat/permissions/RedisPermissionsRepository.java
```diff
@@ -256,17 +303,18 @@ public RedisPermissionsRepository put(@NonNull UserPermission permission) {
             Map<String, String> resourcePermissions = hgetall(timeoutContext, userKey);
             userPermission.addResources(extractResources(resourceType, resourcePermissions));
           });
-      userPermission.setAdmin(redisRead(timeoutContext, c -> c.sismember(adminKey(), id)));
       if (!UNRESTRICTED.equals(id)) {
+        userPermission.setAdmin(redisRead(timeoutContext, c -> c.sismember(adminKey(), id)));
```
cfieber, line 307: micro-optimization assuming the unrestricted user is never admin..

fiat-web/src/main/java/com/netflix/spinnaker/fiat/controllers/AuthorizeController.java
```diff
@@ -307,19 +307,26 @@ public void canCreate(
...
       }
+      log.debug(
+          "Returning fallback permissions (user: {}, accounts: {}, roles: {})",
+          userId,
+          (userPermission != null) ? userPermission.getAccounts() : Collections.emptyList(),
+          (userPermission != null)
+              ? userPermission.getRoles().stream().map(Role::getName).collect(Collectors.toList())
+              : Collections.emptyList());
+    } else {
+      log.debug(
```
cfieber, line 322: One of the cases where we don't fall back happens during gate's auth flow for X509 - this should be more obvious that occurred (userId != authenticatedUserId)

robzienert (approved):

jonsie (approved):

PR #771, opened 2020-09-09
chore(java11): Target Java 11.

plumpy:
After this commit, we can use Java 11 language features and APIs.

I added a single `var` declaration to make it more obvious if any of our various CI systems are misconfigured.


cfieber (approved):

cfieber:
> I added a single var declaration

good idea!

PR #782, opened 2020-09-24
fix(permission): Normalize user permission ID and ensure controllers normalize incoming user IDs

jonsie:
Inputs and outputs normalized so that the user ID lookup is functionally case insensitive.

### DIFF
--- a/fiat-core/src/main/java/com/netflix/spinnaker/fiat/model/UserPermission.java
+++ b/fiat-core/src/main/java/com/netflix/spinnaker/fiat/model/UserPermission.java
@@ -38,6 +38,20 @@ public class UserPermission {
   private Set<Resource> extensionResources = new LinkedHashSet<>();
   private boolean admin = false;
 
+  /**
+   * Custom setter to normalize the input. lombok.accessors.chain is enabled, so setters must return
+   * {@code this}.
+   */
+  public UserPermission setId(String id) {
+    this.id = id.toLowerCase();
+    return this;
+  }
+
+  /** Custom getter to normalize the output. */
+  public String getId() {
+    return this.id.toLowerCase();
+  }
+
   public void addResource(Resource resource) {
     addResources(Collections.singleton(resource));
   }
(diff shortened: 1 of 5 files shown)


fiat-core/src/main/java/com/netflix/spinnaker/fiat/model/UserPermission.java
```diff
@@ -38,6 +38,20 @@
...
+  public UserPermission setId(String id) {
+    this.id = id.toLowerCase();
+    return this;
+  }
+
+  /** Custom getter to normalize the output. */
+  public String getId() {
+    return this.id.toLowerCase();
+  }
+
```
jonsie, line 54: Maybe a bit hacky?  The alternative would be scattering normalize code throughout the permission resolver and repository (and probably other places too).

cfieber: I think this makes sense and is nicely localized

Should we do something at the API to ensure we lowercase request parameters before we start calling things like permission repo? I think a number of those paths just go by userId

cfieber: nevermind .. I see ControllerSupport below!

fiat-roles/src/test/groovy/com/netflix/spinnaker/fiat/permissions/RedisPermissionsRepositorySpec.groovy
```diff
@@ -230,34 +230,34 @@ class RedisPermissionsRepositorySpec extends Specification {
                  .setRoles([role1] as Set))
 
     then:
-    jedis.smembers("unittests:users") == ["testUser"] as Set
-    jedis.smembers("unittests:roles:role1") == ["testUser"] as Set
+    jedis.smembers("unittests:users") == ["testuser"] as Set
```
cfieber, line 233: slight concern over what this might mean for existing data in redis but I think our fallback behaviors will handle it -- anonymous is already lower case so someone that was indexed in mixed case will trip our fallback to re-resolve their permission set

cfieber (approved):

fiat-roles/src/test/groovy/com/netflix/spinnaker/fiat/permissions/RedisPermissionsRepositorySpec.groovy
jonsie: Good callout.  Yeah I think the `PermissionsResolver` will adequately handle the fallback scenario.  Will keep an eye on this as it rolls out though.

PR #785, opened 2020-09-25
fix(errors): Import kork-web ErrorConfiguration into fiat-api configuration, wire up FiatAccessDeniedExceptionHandler

jonsie:


fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatAuthenticationConfig.java
```diff
@@ -83,6 +90,13 @@ FiatWebSecurityConfigurerAdapter fiatSecurityConfig(FiatStatus fiatStatus) {
     return new FiatWebSecurityConfigurerAdapter(fiatStatus);
   }
 
+  @Bean
+  @Order(HIGHEST_PRECEDENCE)
```
jonsie, line 94: This will ensure `FiatAccessDeniedExceptionHandler` always handles access denied errors since kork-web's `GenericExceptionHandlers` is also wired up.

cfieber (approved):
lgtm

PR #804, opened 2020-11-03
Allow to use any ldap attribute as user id for batch role sync

krasilnikov-dmitriy:
Currently we're have two options for batch sync users roles:

1. Sync roles for each user by making request to ldap with **userSearchFilter**. That way allow us to use any ldap attribute as **user id** but it very slow ~ 16min for 1000 users.

2. Sync users roles by making 1 request to ldap using ** groupSearchBase** #573. For that case all user related information gets from member attribute which consists of only DN(distinguished name) because ldap doesn't support join queries. That way dramatically faster ~ 14sec for 1000 users, but we're couldn't use sAMAccountName or any other ldap attributes which we're used to use in **userSearchFilter** and have to use only CN as user id.

Usually ldap contains a lot of users and few roles and in that case we could change our approach for sync users roles.
First of all we can get all users with role by one ldap call like that `(&(objectCategory=user)(memberOf={ROLE_DN}))`. That query return all objects which takes part of the group, but we're allowed to use only strong equality by DN for a memberOf filter.
As a result we'll make N+1 requests to ldap where N is number of groups and it seems to be fast also we could use the same ldap attribute as user id which we using in **userSearchFilter** for syncing single user. In our case it takes 7sec for syncing 250 users in 2 roles.

I understand that request contains breaking changes but maintains 3 different ways to sync roles doesn't sounds like good idea and such approach looks like best compromise: it still fast for large ldap instalation and not so slow for small instalations.

Linked issue: https://github.com/spinnaker/spinnaker/issues/6150


blaquezzz (approved):

fiat-ldap/src/main/java/com/netflix/spinnaker/fiat/config/LdapConfig.java
```diff
@@ -55,10 +55,8 @@ SpringSecurityLdapTemplate springSecurityLdapTemplate() throws Exception {
     MessageFormat userDnPattern = new MessageFormat("uid={0},ou=users");
     String userSearchBase = "";
     String userSearchFilter;
+    String userIdAttribute;
     String groupSearchFilter = "(uniqueMember={0})";
     String groupRoleAttributes = "cn";
-    String groupUserAttributes = "";
```
jasonmcintosh, line 60: SO ... ONE thought, changing the attribute name in this way... vs deprecated handling of it.  NOMINALLY should have a better path to allow migration vs. a complete breaking change, while trying to keep some of the improved functionality under a different path.  Looking at some of this think this would definitely improve group handling but I think we need to provide a migration path.

fieldju:
I haven't actually peer reviewed the code, but I wanted to add a general comment on breaking changes.

We can't just abruptly introduce breaking changes to core functionality, as this produces a very unstable user experience for Spinnakers enterprise customers.

The general approach i'd like to see for something like this is that it is opt in for x releases, then it becomes the default for y releases but you can opt into the legacy provider and announce after z releases you will be deleting the legacy impl.

So i'd like to see the UX for enabling this be like so

`fiat.ldap.rolesProvider: v2` but it defaults to v1 if that is not present.

this might involve making a new module in this project `fiat-ldap-v2` and doing a slight tweak on the v1 config so that it only loads if its set to `v1` and in the base config set `fiat.ldap.rolesProvider: v1`

:point_up: something like this allows us to responsibly deprecate and replace functionality.

krasilnikov-dmitriy:
@fieldju Hi, i've moved all changes to different module could you have a look to it.

Also i've changed the way how to get users from ldap by one query. Now i've construct one query joining the userSearchFilter for all users by | (or operator). As i known ldap support query size up to 10Mb and sounds like it should works for any condition.

fiat-ldap/src/test/groovy/com/netflix/spinnaker/fiat/roles/ldap/LdapUserRolesProviderTest.groovy
```diff
@@ -86,61 +86,6 @@ class LdapUserRolesProviderTest extends Specification {
     "notEmpty"     |_
   }
 
-  void "multiLoadRoles should use loadRoles when groupUserAttributes is empty"() {
```
th3morg, line 89: @krasilnikov-dmitriy Can you please comment on why these test are removed and also why tests for LdapUserRolesProviderV2 have not been added? We aim to add as much test coverage as possible when adding new features, and fixing issues for that matter. Thank you!

krasilnikov-dmitriy: > Can you please comment on why these test are removed
Oh, sorry. Thank you thats my mistake. I revert that that tests.

> why tests for LdapUserRolesProviderV2 have not been added
I think it would be better to get some approvals for that changes (i mean thats i do something right), before invest more time in that changes. Currently that pull request is just proof of concept of the new way how to sync roles.

fiat-ldap-v2/src/main/java/com/netflix/spinnaker/fiat/roles/ldap/LdapUserRolesProviderV2.java
```diff
@@ -0,0 +1,212 @@
...
+
+    // Copied from org.springframework.security.ldap.userdetails.DefaultLdapAuthoritiesPopulator.
+    Set<String> userRoles =
+        ldapTemplate.searchForSingleAttributeValues(
+            configProps.getGroupSearchBase(),
+            configProps.getGroupSearchFilter(),
+            params,
+            configProps.getGroupRoleAttributes());
+
+    log.debug("Got roles for user " + userId + ": " + userRoles);
```
th3morg, line 106: We wrapped line 84 in an if check, should we do the same here?

settings.gradle
```diff
@@ -43,7 +43,8 @@ include 'fiat-api',
         'fiat-google-groups',
         'fiat-ldap',
         'fiat-roles',
-        'fiat-web'
+        'fiat-web',
+        'fiat-ldap-v2'
```
th3morg, line 47: Pretty minor, but I would keep this alphabetized by inserting this line beneath `'fiat-ldap',`

fiat-ldap-v2/src/main/java/com/netflix/spinnaker/fiat/roles/ldap/LdapUserRolesProviderV2.java
krasilnikov-dmitriy: I think yes... and also i wrap the same line in v1 module

fiat-ldap/src/main/java/com/netflix/spinnaker/fiat/roles/ldap/LdapUserRolesProvider.java
```diff
@@ -94,7 +94,9 @@
             params,
             configProps.getGroupRoleAttributes());
 
-    log.debug("Got roles for user " + userId + ": " + userRoles);
+    if (log.isDebugEnabled()) {
+      log.debug("Got roles for user " + userId + ": " + userRoles);
```
jasonmcintosh, line 98: I'd think the if check here for debug is redundant, as log levels would control whether this is output or not?

fiat-ldap-v2/src/main/java/com/netflix/spinnaker/fiat/roles/ldap/LdapUserRolesProviderV2.java
jasonmcintosh: isn't the "if debug enabled" redundant?  e.g. I'd remove the if check from both as the logging service should handle that....

fiat-ldap/src/test/groovy/com/netflix/spinnaker/fiat/roles/ldap/LdapUserRolesProviderTest.groovy
krasilnikov-dmitriy: Add tests for new functionality

fiat-ldap/src/main/java/com/netflix/spinnaker/fiat/roles/ldap/LdapUserRolesProvider.java
deverton: If you're worried about the performance of the log statement, you'd use something like

```java
       log.debug("Got roles for user {}: {}", userId, userRoles); 
```

Which will at least skip the string construction.

fiat-ldap/src/main/java/com/netflix/spinnaker/fiat/config/LdapConfig.java
krasilnikov-dmitriy: we add v2 provider

fiat-ldap/src/main/java/com/netflix/spinnaker/fiat/roles/ldap/LdapUserRolesProvider.java
krasilnikov-dmitriy: fixed

fiat-ldap-v2/src/main/java/com/netflix/spinnaker/fiat/roles/ldap/LdapUserRolesProviderV2.java
```diff
@@ -0,0 +1,202 @@
...
+  @Override
+  public Map<String, Collection<Role>> multiLoadRoles(Collection<ExternalUser> users) {
+    if (StringUtils.isEmpty(configProps.getGroupSearchBase())) {
+      return new HashMap<>();
+    }
+    StringBuilder filter = new StringBuilder();
+    filter.append("(|");
+    users.forEach(
+        u -> filter.append(MessageFormat.format(configProps.getUserSearchFilter(), u.getId())));
+    filter.append(")");
```
jvz, line 142: You can create a new `OrFilter` and run `filter.or(Filter)` on each filter to or together.

PR #818, opened 2021-01-14
fix(roles): ensure usernames are case-insensitively deduped in RedisPermissionsRepository.getAllByRoles()

ajbw:
At the username dedupe stage, pre-emptively call `toLowerCase()` to ensure that the `dedupedUsernames` set has been case-insensitively deduped.

In our environment some users have mixed-case usernames, which would lead to a "deduped" username set containing (for example) both `jdoe` and `JDoe`.  In this situation, `RedisPermissionsRepository.getAllByRoles()` would crash with:

```
    java.lang.IllegalStateException: Duplicate key jdoe (attempted merging values UserPermission(id=jdoe, accounts=[Account(resourceType=account, name=someaccount, cloudProvider=null, permissions=Permissions(permissions={}))], applications=[], serviceAccounts=[], roles=[Role(resourceType=role, name=somerole, source=null)], buildServices=[], extensionResources=[], admin=false) and UserPermission(id=jdoe, accounts=[Account(resourceType=account, name=someaccount, cloudProvider=null, permissions=Permissions(permissions={}))], applications=[], serviceAccounts=[], roles=[Role(resourceType=role, name=somerole, source=null)], buildServices=[], extensionResources=[], admin=false))
```

This is because the username dedupe in `getAllByRoles()` was, prior to this commit, case-sensitive, but the keys for the subsequent `collect()` are converted to lowercase by way of `UserPermission::getId`, leading to key duplication.

By calling `toLowerCase()` to case-insensitively dedupe the username set, the resultant `Map` contains only one (lowercased) entry for each user, as expected.

### DIFF
--- a/fiat-roles/src/main/java/com/netflix/spinnaker/fiat/permissions/RedisPermissionsRepository.java
+++ b/fiat-roles/src/main/java/com/netflix/spinnaker/fiat/permissions/RedisPermissionsRepository.java
@@ -356,7 +356,8 @@ public class RedisPermissionsRepository implements PermissionsRepository {
 
     final Set<String> dedupedUsernames = new HashSet<>();
     for (String role : new HashSet<>(anyRoles)) {
-      dedupedUsernames.addAll(scanSet(roleKey(role)));
+      dedupedUsernames.addAll(
+          scanSet(roleKey(role)).stream().map(String::toLowerCase).collect(Collectors.toSet()));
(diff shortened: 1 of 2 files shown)


jonsie (approved):
Makes sense.  Thanks for fixing this.

link108:
@mergifyio backport release-1.24.x

PR #823, opened 2021-01-25
fix(roles/sync): Fix duplicate key error during sync

adriengentil:
Since https://github.com/spinnaker/fiat/pull/782, our sync process fails with
`java.lang.IllegalStateException: Duplicate key` because we have userIds that are
present multiple times with a different case.


adriengentil:
Already tackled by #818 :facepalm:

PR #828, opened 2021-02-05
fix(redis): Use byte[] instead of String

deverton:
Change the `RedisPermissionsRepository` to use `byte[]` instead of `String` when communicating with Redis. This reduces memory usage with larger permission sets.


robzienert (approved):
Nice!

th3morg:
@deverton It looks like this has approval, but conflicts need to be resolved before it can be merged.

deverton:
@th3morg should be good to go now.

Nessex:
FYI: I've added some rough performance measurement data that is relevant to this PR over here: 

https://github.com/spinnaker/fiat/pull/841#issuecomment-792738769

This change took a whole minute off a 6+ minute sync 👍

cfieber (approved):

PR #829, opened 2021-02-05
fix(roles): Annotate `ServiceAccountPredicateProvider` with `@NonNull`

ajordens:


ajordens:
Matches the rest of `fiat`.

PR #830, opened 2021-02-08
fix(redis): stream read from Redis

deverton:
This change makes the role sync process stream and process the data from Redis incrementally. Previously it would cache the entire data set before processing which would cause `fiat` to consume a large amount of heap on some data sets.

May fix spinnaker/spinnaker#6328 and fix spinnaker/spinnaker#6319

### DIFF
--- a/fiat-roles/src/main/java/com/netflix/spinnaker/fiat/permissions/RedisPermissionsRepository.java
+++ b/fiat-roles/src/main/java/com/netflix/spinnaker/fiat/permissions/RedisPermissionsRepository.java
@@ -18,13 +18,8 @@ package com.netflix.spinnaker.fiat.permissions;
 
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
-import com.fasterxml.jackson.databind.util.ArrayIterator;
 import com.github.benmanes.caffeine.cache.Caffeine;
 import com.github.benmanes.caffeine.cache.LoadingCache;
-import com.google.common.collect.ArrayTable;
-import com.google.common.collect.HashBasedTable;
-import com.google.common.collect.Lists;
-import com.google.common.collect.Table;
 import com.netflix.spinnaker.fiat.config.UnrestrictedResourceConfig;
 import com.netflix.spinnaker.fiat.model.UserPermission;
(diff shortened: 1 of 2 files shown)


fiat-roles/src/main/java/com/netflix/spinnaker/fiat/permissions/RedisPermissionsRepository.java
```diff
@@ -398,47 +389,19 @@ private UserPermission getUserPermission(String userId, RawUserPermission raw) {
...
-              }
-              p.sync();
-            });
-      }
-      return responseTable;
-    } catch (Exception e) {
-      log.error("Storage exception reading all entries.", e);
-    }
-    return null;
+  private RawUserPermission getRawUserPermissionsFromRedis(String userId) {
```
cfieber, line 392: There's a lot of overlap between this and getFromRedis that it would be nice to consolidate

cfieber:
I like the approach - had a comment about the overlap of `getFromRedis` and `getRawPermissionFromRedis` - I'd like to see there only be one of those.

Some of the things in `getFromRedis` are intentionally not pipelined to avoid executing large batches of commands at once against redis and just issuing commands at a time wrapped with `redisRead`. That is an intentional tradeoff of speed vs throughput by trying to spread request processing on redis that I think is worthwhile to keep.

deverton:
In testing so far we've not seen any problems with Redis (well, Elasticache) keeping up with the command rate but I understand the point.

Combining the two functions would be ideal, but then what time out would apply during a sync process? Do you timeout the whole sync? Or just the individual value being synced at the time?

deverton:
Updated the PR to unify the methods. This deletes a lot of code so I'm a bit suspicious about it still being correct, but in the testing I've done it seems fine. There's no more batching of user fetches though which seems like it should slow things down but doesn't seem to. It does open up the possibility of using `parallelStream()` and a fork-join pool to sync users in parallel though.

robzienert (approved):

deverton:
To follow up, we've been running this change in production for a while now and the memory reduction is pretty reasonable though it really only affects the peak usage.

|                   | Average | Max |
|-------------|---------|------|
| Before  | 8 GB | 22 GB |
| After    | 8 GB | 15 GB |

PR #832, opened 2021-02-10
fix(web): Override the default `TaskScheduler`

ajordens:
`fiat` has multiple `@Scheduled` functions and the default single-threaded
task execution likely isn't able to keep up.

This PR changes the default to `5` but can be overriden with:

```
fiat:
  scheduler:
    poolSize: 10
```


jonsie (approved):
Makes sense, not sure why the build is failing though.

ajordens:
@jonsie Our friendly spotless that rarely kicks in for me when I'm working via the IDE.

luispollo:
Nice find! 👍

PR #838, opened 2021-02-26
feat(sql): Add an SQL permission repository

deverton:
This PR maintains backwards compatibility and does not modify existing controllers.

Implements a permission repository backed by SQL with both MySQL and Postgres supported.

Enabling SQL as the permission repository is done as follows:

```yaml
sql:
  enabled: true
  connectionPools:
    default:
      jdbcUrl: jdbc:mysql://.....
      user: ...
      password: ...
      default: true
  migration:
      jdbcUrl: jdbc:mysql://.....
      user: ...
      password: ...

permissionsRepository:
  redis:
    enabled: false
  sql:
    enabled: true
```

Integration tests have been updated to enable testing with the SQL provider (using testcontainers), but still default to Redis only as that's faster.

Depends on #835 a bit for testcontainer usage.


fiat-core/src/main/java/com/netflix/spinnaker/fiat/model/UserPermission.java
```diff
@@ -53,30 +53,27 @@ public String getId() {
...
-            serviceAccounts.add((ServiceAccount) resource);
-          } else if (resource instanceof Role) {
-            roles.add((Role) resource);
-          } else if (resource instanceof BuildService) {
-            buildServices.add((BuildService) resource);
-          } else {
-            extensionResources.add(resource);
-          }
-        });
+    resources.forEach(this::addResource);
```
robzienert, line 76: 👍  Good improvement.

robzienert (approved):
I'm no fiat expert, but what you have here makes sense to me from a general technical perspective. Will defer to @jonsie and @cfieber for actual implementation bits.

Have you built and validated this change in any environment (e.g. load tests)?

Great work, thanks for taking this on.

deverton:
It's been subjected to some load testing in our staging environment. For the workload there

deverton:
Once I get the data migration logic in place we'll try rolling this in our production environment soonish.

jonsie (approved):
LGTM.  Very curious to see how this performs in your production environment.  I may take a shot at running this in our preprod env too.

deverton:
Just ran in to an issue with how the dual repository looks up the beans from the context which I'm hoping to fix today.

jonsie:
@deverton Fiat should rebuild all permissions on boot (and on a schedule) so I'm not sure the dual repository is necessary anyways.

deverton:
It might be how we've got things configured, but new instances of Fiat don't seem to role sync immediately. There's typically roughly 10 minutes before the first scheduled sync runs and there's only a single instance in this case.

```json
{"@timestamp":"2021-03-03T04:45:02.602+00:00","@version":1,"message":"Server is now HEALTHY. Hooray!","logger_name":"com.netflix.spinnaker.fiat.config.ResourceProvidersHealthIndicator","thread_name":"http-nio-0.0.0.0-7003-exec-2","level":"INFO","level_value":20000}
...
{"@timestamp":"2021-03-03T04:53:27.007+00:00","@version":1,"message":"Acquired Lock Lock{name='fiat.userrolessyncer', ownerName='spin-fiat-6954658dd9-nrhmj', leaseDurationMillis=10000, successIntervalMillis=600000, failureIntervalMillis=600000, version=1614747207004, ownerSystemTimestamp=1614747207004, attributes=''}.","logger_name":"com.netflix.spinnaker.kork.jedis.lock.RedisLockManager","thread_name":"scheduler-3","level":"INFO","level_value":20000}
```

deverton:
I've dropped the dual repository code out of this PR for now. I'll do some more testing locally to see if the issues on cutover persist for us.

Would you like me to squash these commits up before merge?

jonsie:
Yeah, 10 minutes is the default of `syncDelayMs` so perhaps you are using the default here? We have that dropped down to 2 minutes in our config.

No need to squash commits, I'll squash them on merge.

deverton:
@jonsie so rolled this out in production and it seems to be working fine. Performance for bulk operations like `getAllById` and `getAllByRole` are much faster. The `get()` operation is also faster for our data set. The big downside is that `put()` is much, much slower, nearly 10x as much (50 ms vs 500 ms). This means that the user role sync process is still faster than when we were on Redis but is still too slow.

The `put()` operation was implemented pretty naively by just deleting and re-inserting records so there's obviously room for improvement there.

jonsie:
@deverton Yeah I was wondering about that delete/insert in the put operation but I decided not to say anything since it's +/- the same behavior in the Redis permission repository.

I think all in all this is a good sign though.  We're running this in our test environment right now and things look good, I will promote this to our staging environment soon.

PR #839, opened 2021-03-02
refactor(ldap/logging): Add userId in error log

kowshikRoy:
When userId is not found in LDAP, it just shows the message `Unable to find a single user entry` but not mentions which `userId` has caused this error. Having this userId will enable teams to troubleshoot the issue quickly.


fiat-ldap/src/main/java/com/netflix/spinnaker/fiat/roles/ldap/LdapUserRolesProvider.java
```diff
@@ -166,7 +166,7 @@ private String getUserFullDn(String userId) {
                 configProps.getUserSearchBase(), configProps.getUserSearchFilter(), formatArgs);
         partialUserDn = res.getDn().toString();
       } catch (IncorrectResultSizeDataAccessException e) {
-        log.error("Unable to find a single user entry", e);
+        log.error("Unable to find a single user entry for " + userId, e);
```
deverton, line 169: Since this is using [slf4j](http://www.slf4j.org/faq.html#logging_performance) this should be

```java
log.error("Unable to find a single user entry for {}", e);
```

deverton (changes requested):

fiat-ldap/src/main/java/com/netflix/spinnaker/fiat/roles/ldap/LdapUserRolesProvider.java
kowshikRoy: error method has overloaded function definition as bellow. I am using that in this case.
```
  Log an exception (throwable) at the ERROR level with an accompanying message.
  Params:
    msg – the message accompanying the exception
    t – the exception (throwable) to log
public void error(String msg, Throwable t);
```

deverton:
It's not obvious from the API, but the last parameter in the parameter list is treated as a `Throwable` for logging purposes. See http://www.slf4j.org/faq.html#paramException

deverton (approved):

mattgogerly (approved):

PR #841, opened 2021-03-03
perf(redis): Performance improvements for RedisPermissionsRepository

Nessex:
## WHAT

This PR makes a number of changes to improve the performance of RedisPermissionsRepository. As these are breaking changes from the perspective of the redis "schema" and other versions of Fiat, the changes are all together rather than in separate PRs. Splitting it up will require further renaming of redis keys to prevent conflicts, and this would further bloat redis memory usage. These changes are much better together than apart!

Summary:
 - Removes all uses of the O(n) operations [`HMSET`](https://redis.io/commands/hmset) / [`HGETALL`](https://redis.io/commands/hgetall) as the individual map keys are never used anyway. Replaces them with the O(1) [`GET`](https://redis.io/commands/get) / [`SET`](https://redis.io/commands/set) operations.
 - Serializes and deserializes objects in one go, no more strings containing JSON inside JSON which need to be parsed separately
 - Adds fast LZ4 deflate for compressing redis payloads, this has negligible CPU impact in exchange for significantly less IO. IO makes up a significant amount of time in `/roles/sync`
 - Adds a new permissions key `permissions-v2` which is used instead of `permissions` for the above changes. That means updating fiat will not have any conflicts with old keys that hold redis hashmap data. Downside: this means that after the upgrade, slightly more redis memory will be used for the new data + old data. This can be cleaned up by deleting redis keys under `spinnaker:fiat:permissions:*`

## WHY

The current RedisPermissionRepository is too slow for our use-case. `/roles/sync`, and transitively all the operations that trigger it including saving pipelines, creating applications etc. are taking multiple minutes and the UI is responding as if the operation timed out. This was an attempt (though it's not yet sufficient for us) to make fiat fast enough.

## RESULTS

[See below](https://github.com/spinnaker/fiat/pull/841#issuecomment-792738769)

## ~TODO~ <- All done

These are mostly administrative.

 - [x] ~Take a second look at splitting up the PR~ This would require multiple data migrations.
 - [x] Get internal code review
 - [x] Squash changes to make them follow commit conventions
 - [x] Add master + https://github.com/spinnaker/fiat/pull/828 to performance comparison
 - [x] Wait for the [`Use byte[] instead of String` PR](https://github.com/spinnaker/fiat/pull/828) to be merged
 - [x] Resolve conflicts after that PR is merged


Nessex:
# Some Data

This fiat deployment has roughly 2500 service accounts, most of which have the exact same role. This replicates the situation we are trying to resolve whereby certain operations sporadically fail because of the triggered `/roles/sync` operation taking too long.

Please note that some of the performance differences can also be attributed to https://github.com/spinnaker/fiat/pull/830 and https://github.com/spinnaker/fiat/pull/828 which are included in the after image. I'll look into a comparison with those PRs in the next few days.

| before | after |
| --- | --- |
| fiat 1.24.1 | this PR |

## Fiat syncUsers Time (max)

| before | after |
| --- | --- |
| Screen Shot 2021-02-24 at 22 06 07 | Screen Shot 2021-02-24 at 22 06 57 |

## Fiat syncTime (max)

| before | after |
| --- | --- |
| Screen Shot 2021-02-24 at 22 06 15 | Screen Shot 2021-02-24 at 22 06 50 |

## Fiat CPU usage

The purple spike was a new pod starting with the official 1.24.1 image, not related.

| before | after |
| --- | --- |
| Screen Shot 2021-02-24 at 22 10 26 | Screen Shot 2021-02-24 at 22 09 27 |

## Fiat Memory Usage

This is hard to interpret as the memory usage is always climbing up to the high memory limit I had set. The GC isn't working too hard here. The before image container was also running for significantly less time, usually memory just keeps climbing on average until the memory limits are hit.

If you look closely, you'll see that the dashed line, committed memory, is also much higher before the change.

At the lowest points (after a major GC cleanup), the before image drops to about 900M, and the after image drops to about 750M.

| before | after |
| --- | --- |
| Screen Shot 2021-02-24 at 22 10 17 | Screen Shot 2021-02-24 at 22 09 13 |

## Redis commands total time

| before | after |
| --- | --- |
| Screen Shot 2021-02-24 at 22 11 02 | Screen Shot 2021-02-24 at 22 08 48 |

## Redis network traffic

A bit useless without averaging out the line rate, but you can see that it's (if I'm a bit generous) an order of magnitude less traffic.

| before | after |
| --- | --- |
| Screen Shot 2021-02-24 at 22 10 39 | Screen Shot 2021-02-24 at 22 08 25 |

## Redis microseconds per call

This doesn't say much, other than that HMSET is expensive. The scale of the graph changes a lot without it.

| before | after |
| --- | --- |
| Screen Shot 2021-02-24 at 22 11 26 | Screen Shot 2021-02-24 at 22 27 45 |

Nessex:
# Some More Data

This fiat deployment has roughly 2500 service accounts, most of which have the exact same role. This replicates the situation we are trying to resolve whereby certain operations sporadically fail because of the triggered /roles/sync operation taking too long.

I have about an hour of sync data for each of the following three versions of Fiat. deverton:redis-bytes is included because this change builds upon that PR which hasn't been merged yet. By including it, you can see how much of the performance impact can be attributed to each change. The master and deverton:redis-bytes commits were the latest commits in each branch at the time of comment.

| deployment | version |
| --- | --- |
| 6987b688f7 | This PR (commit ded494b) |
| 869778d568 | spinnaker:master-1be70bacd36 |
| 55ff6c6774 | deverton:redis-bytes-967c25a |

### syncUsers Time (max, seconds)

This is the time spent syncing non-anonymous user roles.

| master | deverton:redis-bytes | this PR |
| --- | --- | --- |
| Screen Shot 2021-03-08 at 21 24 08 | Screen Shot 2021-03-08 at 19 59 32 | Screen Shot 2021-03-08 at 15 49 48 |

### syncTime (max, seconds)

This is roughly the total time of the user roles sync.

| master | deverton:redis-bytes | this PR |
| --- | --- | --- |
| Screen Shot 2021-03-08 at 21 24 00 | Screen Shot 2021-03-08 at 19 59 25 | Screen Shot 2021-03-08 at 15 49 32 |

### JVM Memory pre-GC

This is hard to interpret as the memory usage is always climbing up to the high memory limit I had set. The GC isn't working too hard here. The before image container was also running for significantly less time, usually memory just keeps climbing on average until the memory limits are hit.

This screenshot shows a representative point pre-GC event (some time during the user role sync process, when a lot of memory is being used).

| master | deverton:redis-bytes | this PR |
| --- | --- | --- |
| Screen Shot 2021-03-08 at 21 25 03 | Screen Shot 2021-03-08 at 20 00 10 | Screen Shot 2021-03-08 at 15 50 45 |

### JVM Memory post-GC

As above, however this screenshot shows a representative point post-GC event (some time not long after the user role sync ends).

| master | deverton:redis-bytes | this PR |
| --- | --- | --- |
| Screen Shot 2021-03-08 at 21 40 09 | Screen Shot 2021-03-08 at 20 00 04 | Screen Shot 2021-03-08 at 15 50 39 |

### Fiat Network IO

The area under the graph is most important here.

| master | deverton:redis-bytes | this PR |
| --- | --- | --- |
| Screen Shot 2021-03-08 at 21 24 17 | Screen Shot 2021-03-08 at 20 00 16 | Screen Shot 2021-03-08 at 15 51 19 |

### Fiat CPU Usage

The area under the graph is most important here.

| master | deverton:redis-bytes | this PR |
| --- | --- | --- |
| Screen Shot 2021-03-08 at 21 23 33 | Screen Shot 2021-03-08 at 19 59 55 | Screen Shot 2021-03-08 at 15 51 08 |

### Redis CPU Usage

Watch out for the significant scale change in the last image.

| master | deverton:redis-bytes | this PR |
| --- | --- | --- |
| Screen Shot 2021-03-08 at 21 41 58 | Screen Shot 2021-03-08 at 20 00 30 | Screen Shot 2021-03-08 at 15 51 50 |

### Redis Command Calls

Each color is a different command type (though Datadog saw fit to make two commands the exact same shade of purple). The area under the graph is most important. Not a huge change, but there is definitely a reduction overall.

| master | deverton:redis-bytes | this PR |
| --- | --- | --- |
| Screen Shot 2021-03-08 at 21 42 13 | Screen Shot 2021-03-08 at 20 00 49 | Screen Shot 2021-03-08 at 15 52 34 |

### Redis Commands Total Time

Stacked areas, each color represents a different command type. Replacing HMSET / HGETALL with the cheaper SET / GET makes a significant difference.

| master | deverton:redis-bytes | this PR |
| --- | --- | --- |
| Screen Shot 2021-03-08 at 21 42 20 | Screen Shot 2021-03-08 at 20 00 36 | Screen Shot 2021-03-08 at 15 52 27 |

### Redis Network Traffic

The area under the graph is most important.

| master | deverton:redis-bytes | this PR |
| --- | --- | --- |
| Screen Shot 2021-03-08 at 21 42 32 | Screen Shot 2021-03-08 at 20 00 56 | Screen Shot 2021-03-08 at 15 52 05 |

### Redis Microseconds Per Call

This is the average number of microseconds spent per call of each redis command (GET, SET, HMSET etc.). It doesn't say much, other than that HMSET is expensive. The scale of the graph changes a lot without it.

Note: The sudden spikes in the `this PR` image are from the `CONFIG` command, which is not related to this change.

| master | deverton:redis-bytes | this PR |
| --- | --- | --- |
| Screen Shot 2021-03-08 at 21 42 49 | Screen Shot 2021-03-08 at 20 01 04 | Screen Shot 2021-03-08 at 15 52 54 |

Nessex:
## Post-apply redis bloat

One downside to this PR is that it uses new redis keys for permissions, leading to some bloat. I tested the upgrade with a larger fiat deployment and saw the following increase in memory usage when upgrading:

| before | after |
| --- | --- |
| Screen Shot 2021-03-09 at 12 03 00 | Screen Shot 2021-03-09 at 12 03 13 |

It's less than I expected. This deployment has a few thousand service accounts and a few hundred applications.

deverton:
You may also be interested in https://github.com/spinnaker/orca/pull/4083 which we use to cut down on the number of distinct service accounts.

Nessex:
@deverton thanks! We're working on testing out that change as well. We have lots of users with lots of different roles though, so in our case we will likely need both + some further changes at the UserRolesSyncer level.

deverton:
Yeah, I think we're likely going to need some refactoring of the sync process and quite possibly the data layout to really scale Fiat up to how we use it.

With the shared managed service account patch, we found that it helped cut down the number of accounts a lot. Before we applied it we hard somewhere around 4000 service accounts and after the number is closer to 200.

deverton:
@Nessex you may be interested in #854 which was a huge win for us in terms of sync time. It's back end independent as well.

Nessex:
> you may be interested in #854 which was a huge win for us in terms of sync time

#854 looks great, and we're so close to fixing our timeout issues at least that it should push us over the edge if we combine both changes. We've been running this current PR for quite a while now and it's improved things significantly, so I hope both changes can eventually be merged to keep redis on equal footing with the SQL repository.

The change pushed above was just to account for a change to the unit tests in master. Otherwise this PR is unchanged.

Nessex:
Unfortunately it looks like #854 caused about a 30% increase in time to sync for our test environment.

Screen Shot 2021-04-15 at 19 35 51

Key:
 - Dark blue / Purple == #841
 - Light Blue / Yellow == #841 + #854

deverton:
Very interesting. That suggests that the change is very dependent on the back end and or data set being processed. I'm surprised that it was slower though given that #854 should be doing less work. Do you have the ability to attach a profiler to the sync process? I'd be very curious to hear what the top methods are in your case.

jasonmcintosh:
@Nessex did you ever figure anything out on your performance issues with the 854 PR?

Nessex:
@jasonmcintosh Not really sorry, haven't had time to come back and look at the performance of that PR. This PR solved the vast majority of our performance issues so it hasn't been a high priority.

deverton:
Would it be worth putting the compression part (and the change in key) behind a configuration option? At least that way people can opt-in to the change and be ready to clean up the old keys.

Nessex:
@deverton Hmm, I think it would need to be a complete parallel implementation of RedisPermissionRepository at that point, the changes run pretty deep. For example, the change from redis hash maps to redis values, and the change to make JSON no longer nested would both also conflict with the old data. That's the reason I originally kept this all together in one changeset, which I understand is less than ideal.

Is there any precedent for a change like this elsewhere in Spinnaker which could be emulated? I'm hesitant to create a RedisPermissionRepository2 (better name TBD) because it'll be a bit of a mess maintenance-wise and probably confusing for users to know which one to pick.

deverton:
There is a migration framework within [front50](https://github.com/spinnaker/front50/tree/master/front50-migrations/src/main/java/com/netflix/spinnaker/front50/migrations) but I don't think it's quite the right fit this.

Thinking on it a bit more, it's probably better that it is under a separate key as it makes rolling back easier since the old data will still be there. It's probably sufficient to document that the old key can be cleaned up at some point after deployment of the new version. The data is ephemeral anyway, so worst case

Possibly a less implementation specific name for the key space might be better as well? Instead of `permissions-lz4` maybe `permissions-v2` which matches how the schema has been versioned on places like `clouddriver`.

Nessex:
> There is a migration framework within front50 but I don't think it's quite the right fit this.

Yeah it's a bit of a weird migration, there's nowhere great in the lifecycle that I can see to run an automated cleanup of old keys that wouldn't require at least handling the combined total memory usage of old and new data anyway. The impact for our case was super small, nothing particularly notable, but it's hard for me to say if our situation would be shared by all or if some edge case would cause this to significantly bloat memory in certain cases.

> It's probably sufficient to document that the old key can be cleaned up at some point after deployment of the new version

Right, if I recall correctly that's as simple as:

```
redis-cli KEYS "spinnaker:fiat:permissions:*" | xargs redis-cli DEL
```

This will also delete the still-valid `spinnaker:fiat:permissions:admin` key, but it is a minor inconvenience as that will be re-created anyway.

> maybe `permissions-v2` which matches how the schema has been versioned on places like `clouddriver`.

Done 👍

deverton (approved):

cfieber (approved):
this aligns with a change I've made internally to precompute the user permission and store it in a single key and we've seen great success from that

lgtm

link108:
@mergifyio backport release-1.27.x

link108:
@mergifyio backport release-1.26.x

PR #843, opened 2021-03-08
fix(*): Optionally configure a UserRoleSyncer lock name

jonsie:
When trying to test the SQL permission repository side-by-side with the Redis permission repository, we find that the lock is biased to the instance that last had it - in this case the lock is biased towards the Redis permission repository instance.  This will allow us to roll out different a lock name for the cluster configured to use the SQL permission repository.


cfieber (approved):

jonsie:
@deverton Related to testing the SQL permission repository.

PR #844, opened 2021-03-09
perf(sql): smarter put operations when using sql

deverton:
This change attempts to detect whether to insert/update/delete
permissions rather than delete everything and re-insert.


jonsie:
@deverton Any data on this change?

deverton:
Haven't had a chance to test it at scale yet. In the smaller test environment, it's a bit slower on average so there's something I'm not understanding yet.

deverton:
Closing this as I think the schema needs refactoring first.

PR #845, opened 2021-03-10
chore(permissions): add bulk put method

deverton:
Add a bulk put method to the `PermissionsRepository` class to allow implementations to optimise bulk puts.

### DIFF
--- a/fiat-roles/src/main/java/com/netflix/spinnaker/fiat/roles/UserRolesSyncer.java
+++ b/fiat-roles/src/main/java/com/netflix/spinnaker/fiat/roles/UserRolesSyncer.java
@@ -67,6 +67,7 @@ public class UserRolesSyncer {
   private final long syncDelayMs;
   private final long syncFailureDelayMs;
   private final long syncDelayTimeoutMs;
+  private final String lockName;
 
   private final Registry registry;
   private final Gauge userRolesSyncCount;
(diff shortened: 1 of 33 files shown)


jonsie (approved):

fiat-roles/src/main/java/com/netflix/spinnaker/fiat/roles/UserRolesSyncer.java
```diff
@@ -264,8 +263,8 @@ public long updateUserPermissions(Map<String, UserPermission> permissionsById) {
         timeIt(
             "syncUsers",
             () -> {
-              Collection<UserPermission> values = permissionsResolver.resolve(extUsers).values();
-              values.forEach(permissionsRepository::put);
+              Map<String, UserPermission> values = permissionsResolver.resolve(extUsers);
+              permissionsRepository.putAllById(values);
```
jonsie, line 267: Good cleanup.

fiat-sql/src/main/kotlin/com/netflix/spinnaker/fiat/permissions/SqlPermissionsRepository.kt
```diff
@@ -110,6 +110,10 @@ class SqlPermissionsRepository(
         return this
     }
 
+    override fun putAllById(permissions: MutableMap<String, UserPermission>?) {
```
jonsie, line 113: Are you planning on mutating this map?  If not, may as we just be a `Map`.

deverton: Good point. I'll change  that since I can't think of a reason to mutate the map there.

PR #848, opened 2021-03-12
chore(sql): add foreign key constraints

deverton:
Add a foreign key constraint from the permission table to the user table.


deverton:
Actually, I think I'm going to have to redo the table model anyway. We end up with about 8 million rows with the current layout and there's no good way to make it performant.

PR #849, opened 2021-03-14
chore(sql): normalize permission schema

deverton:
Normalizes the permissions schema to improve scalability and optimize
code based on the new schema.

The biggest change is that resources are now stored once rather than
duplicated for each user. This means bulk operations are much faster at
the cost of more queries.


deverton:
Sorry for the churn, should have left this in draft until it was ready. I think this code is about as optimal as it can be using SQL and combines a number of techniques to try and go faster. It's still in testing at the moment and we'll work on a production load test soon.

jonsie:
@deverton How are you feeling about this?  Have you been able to test it out in production?

deverton:
@jonsie we have a production test scheduled for Monday. I've been load testing it in our staging environment which is where the last few changes have come from. So far things look good but we'll know more next week.

fiat-sql/src/main/kotlin/com/netflix/spinnaker/fiat/permissions/sql/TableDefinitions.kt
```diff
@@ -17,31 +17,145 @@
...
+            "${ResourceType.APPLICATION}" -> ResourceType.APPLICATION
+            "${ResourceType.BUILD_SERVICE}" -> ResourceType.BUILD_SERVICE
+            "${ResourceType.ROLE}" -> ResourceType.ROLE
+            "${ResourceType.SERVICE_ACCOUNT}" -> ResourceType.SERVICE_ACCOUNT
+            else -> throw IllegalArgumentException("unknown resource type $databaseObject")
+        }
+    }
+
+    override fun to(userObject: ResourceType?): String? {
+        return userObject?.toString()
```
jonsie, line 158: Thoughts on moving breaking this out to a new package like `com.netflix.spinnaker.fiat.permissions.sql.tables`?  And then have separate files for each table and table record?  Might make it a bit easier to parse.

deverton: Should be much nicer now.

fiat-sql/src/main/kotlin/com/netflix/spinnaker/fiat/permissions/SqlPermissionsRepository.kt
```diff
@@ -61,59 +61,83 @@ class SqlPermissionsRepository(
...
+            }
+
+            // Tidy up deleted users and permissions
+            jooq.transactional(sqlRetryProperties.transactions) { ctx ->
+                val batch = mutableListOf<Query>()
+
+                val existingIds = ctx.select(USER.ID).from(USER).fetch(USER.ID).toSet();
+                val toDelete = existingIds.minus(permissions.keys)
 
+                if (toDelete.isNotEmpty()) {
```
jonsie, line 113: Do you know under what conditions these deletes run?

deverton: Hm, I assumed `fiat` needed to do some tidy up in the case that service accounts were deleted. But looking at https://github.com/spinnaker/front50/blob/0887c6b2b7a25b433fff49af6f93ea79a66e173e/front50-core/src/main/java/com/netflix/spinnaker/front50/ServiceAccountsService.java#L80 it looks like `front50` will take care of triggering a sync etc.

The only thing is I think it would leave the user record around in `fiat` so that would still need to be deleted. I'll do some more digging.

jonsie: You might be right, I'm honestly just not sure.  Feels like it doesn't hurt to keep this here, just not totally sure what condition this supports.

The front50 triggered sync is conditional though, so we can't always depend on that.

deverton:
That last commit fixed the one niggle we'd seen during testing where there were more deletes than expected. Looking good for deployment to our production environment on Monday.

deverton:
We attempted to roll this out in production on Monday and it was going well right up until we hit a flood of dead lock exceptions. Still doing RCA on that, but it seems the `put()` and `putAllById()` operations can conflict in a very bad way resulting in huge amount of query churn.

Performance wise it was pretty decent. Comparing a Elasticache `cache.m6g.2xlarge` to a RDS `db.r6g.4xlarge` with about 1500 users and 5000 resources. That works out to roughly 7.5 million permissions.

| Storage | Action | Min | Average | Max |
| -------  | ------- | ----|---------|----|
| Redis | `get` | 552 ms | 699 ms | 1000 ms |
| SQL   | `get` | 163 ms | 213 ms | 812 ms |
| Redis | `put` | 71 ms | 97 ms | 242 ms
| SQL   | `put` | 2000 ms | 5000 ms | 30000 ms |
| Redis | `getAllById` | 2.93 minutes | 5.55 minutes | 9.18 minutes |
| SQL   | `getAllById` | 3 s | 19 s | 37 s |
| Redis | `getAllByRoles` | 3.54 minutes | 4.31 minutes | 3.83 minutes |
| SQL   | `getAllByRoles` | 6.37 minutes | 6.37 minutes | 6.37 minutes |
| Redis | `putAllById` | 18 s | 43 s | 57 s |
| SQL   | `putAllById` | 6 s | 34 s | 75 s |

The caveat to that is when dead locks happen, performance drops for write operations rock due to the retries.

deverton:
Taking this back in to draft since we're still having issues with the implementation at scale. Day-to-day operations are fine, but the bulk operations are still too slow for our use. There's also some major issues during our batch migration process (we're migrating from an in-house Spinnaker equivalent) where the volume of application on-boarding and thus permission updates causes Fiat with SQL to fall over.

deverton:
Current plan is to be even more like clouddriver at split up the reads/writes in to batches.

deverton:
We've been running this in production for the last month (along with #854 ) with no identified bugs. Perf of sync process is still not great but works well enough for our use case.

cfieber (approved):

PR #854, opened 2021-04-09
perf: Reduce Cost Of Sync Process

deverton:
This change reduces the cost of generating the in-memory model for `UserPermission` and can cut sync time substantially for installations with a high number of resources, roles, and users. There should be no compatibility changes.

The primary changes are:

* Pre-compute the hash code for `ResourceType`, `Role`, and `Permission`. These objects are immutable are are treated as such within the code. At scale, the hash computation becomes expensive given the extensive usage of `Map` and `Set`.
* Change the internal representation of the roles associated with an `Authorization` type within `Permission` to be a `Set`. This makes the calculation of effective permissions vastly cheaper as the call to `Collections.disjoint()` can use an optimised path for `Set` rather than iterating over two lists of roles.
* Change how the hash code is computed for the Fiat domain model and only use fields that are relevant, typically excluding the `permissions` object. This reduces the cost of computing the hash which is a benefit to the sync process as it typically calls `UserPermission.getAllResources()` a lot.
* Improve the hash code calculation for Fiat domain to include the resource type value. This reduces the chance of collisions within the `Set` when calling `UserPermission.getAllResources()`

These changes combined and deployed in conjunction with #849 our role sync times go from an average of 4.5 minutes to 1.5 minutes to sync around 21,000,000 permissions. This should also benefit anyone using the Redis back end as the changes are storage layer independent.

### DIFF
--- a/fiat-api/src/test/groovy/com/netflix/spinnaker/fiat/shared/FiatPermissionEvaluatorSpec.groovy
+++ b/fiat-api/src/test/groovy/com/netflix/spinnaker/fiat/shared/FiatPermissionEvaluatorSpec.groovy
@@ -79,7 +79,7 @@ class FiatPermissionEvaluatorSpec extends FiatSharedSpecification {
             applications: [
                 new Application(name: "abc-def",
                                 permissions: Permissions.factory([
-                                    (Authorization.READ): ["testRole"]
+                                    (Authorization.READ): ["testRole"] as Set
                                 ])),
             ],
             roles: [new Role("testRole")]
--- a/fiat-core/src/main/java/com/netflix/spinnaker/fiat/model/resources/Account.java
+++ b/fiat-core/src/main/java/com/netflix/spinnaker/fiat/model/resources/Account.java
@@ -26,7 +26,9 @@ import org.springframework.stereotype.Component;
 
 @Component
 @Data
-@EqualsAndHashCode(callSuper = false)
+@EqualsAndHashCode(
+    callSuper = false,
+    of = {"resourceType", "cloudProvider", "name"})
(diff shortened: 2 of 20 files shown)


jonsie (approved):
I like it.  Curious what @cfieber thinks.

deverton:
This chart summarises the performance changes. The first part is Redis backed (and mostly just failing when the lock heart beat times out). The middle part is with the SQL backend and the final part is with this patch.

image

fieldju:
@cfieber do you have any input on this PR.
I'd like to get this merged soon, I've got people asking for it.

deverton:
@jonsie @cfieber we've been running this for over a month in production along with #849 with no issues and greatly improved performance for our use case. Would you consider approving this to merge for 1.27 since I think that's getting tagged soon?

mattgogerly:
Any updates on this @deverton?

deverton:
@mattgogerly nothing from my end. We're still running with this patch in production and no issues identified.

cfieber (approved):

jasonmcintosh:
@deverton Has there been any work/thoughts to address https://github.com/spinnaker/fiat/pull/841#issuecomment-820322960 ?  My concern on merging would be breaking existing REDIS users (which is a large segment of the user base at this point in time).

deverton:
@jasonmcintosh I haven't looked at it since we don't use Redis for Fiat anymore. Ideally we'd need someone to profile the patch with Redis to see where the hot spots have moved to.

My guess is that there's something about the data that results in far more hash collisions, but even then I wouldn't expect it to be slower since in the worst case it would result in the same lookup cost as using a List.

PR #879, opened 2021-09-09
fix: improve performance of role sync with SQL

deverton:
The permission rebuild process during role synchronization only requires
the set of roles that each user has. This change modifies the SQL and
Redis `PermissionRepository` instances to only return the roles.

For our use case this drops about 2 minutes of the sync time (so from 3 minutes to 1 minute).

### DIFF
--- a/fiat-roles/src/main/java/com/netflix/spinnaker/fiat/permissions/PermissionsRepository.java
+++ b/fiat-roles/src/main/java/com/netflix/spinnaker/fiat/permissions/PermissionsRepository.java
@@ -60,12 +60,13 @@ public interface PermissionsRepository {
   Map<String, Set<Role>> getAllById();
 
   /**
-   * Gets all Roles in the repository that has at least 1 of the specified roles, keyed by user ID.
-   * Because this method is usually used in conjunction with updating/syncing the users in question,
-   * the returned map will also contain the unrestricted user. If anyRoles is null, returns the same
-   * result as getAllById() (which includes the unrestricted user). If anyRoles is empty, this is an
-   * indication to sync only the anonymous/unrestricted user. When this is the case, this method
-   * returns a map with a single entry for the unrestricted user.
+   * Gets a map of all users and their Roles from the repository where the user has at least one of
+   * the specified named roles. Because this method is usually used in conjunction with
+   * updating/syncing the users in question, the returned map will also contain the unrestricted
+   * user. If anyRoles is null, returns the same result as getAllById() (which includes the
+   * unrestricted user). If anyRoles is empty, this is an indication to sync only the
+   * anonymous/unrestricted user. When this is the case, this method returns a map with a single
+   * entry for the unrestricted user.
    *
    * @param anyRoles
    * @return
--- a/fiat-roles/src/main/java/com/netflix/spinnaker/fiat/permissions/PermissionsRepository.java
+++ b/fiat-roles/src/main/java/com/netflix/spinnaker/fiat/permissions/PermissionsRepository.java
@@ -17,9 +17,11 @@
 package com.netflix.spinnaker.fiat.permissions;
 
 import com.netflix.spinnaker.fiat.model.UserPermission;
+import com.netflix.spinnaker.fiat.model.resources.Role;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.Set;
 
 /**
  * A PermissionsRepository is responsible for persisting UserPermission objects under a user ID key.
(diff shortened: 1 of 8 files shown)


deverton:
@cfieber not sure if this is what you had in mind but it's a pretty big perf win so far.

fiat-roles/src/main/java/com/netflix/spinnaker/fiat/permissions/PermissionsRepository.java
```diff
@@ -54,21 +56,21 @@
...
+  Map<String, Set<Role>> getAllById();
 
   /**
-   * Gets all UserPermissions in the repository that has at least 1 of the specified roles, keyed by
-   * user ID. Because this method is usually used in conjuction with updating/syncing the users in
-   * question, the returned map will also contain the unrestricted user. If anyRoles is null,
-   * returns the same result as getAllById() (which includes the unrestricted user). If anyRoles is
-   * empty, this is an indication to sync only the anonymous/unrestricted user. When this is the
-   * case, this method returns a map with a single entry for the unrestricted user.
+   * Gets all Roles in the repository that has at least 1 of the specified roles, keyed by user ID.
```
mattgogerly, line 63: As someone largely unfamiliar with the inner workings of Fiat persistence this doesn't really make sense? Why does a `Role` object have one of the specified roles - is a Role not just a single role?

fiat-sql/src/main/kotlin/com/netflix/spinnaker/fiat/permissions/SqlPermissionsRepository.kt
```diff
@@ -118,23 +119,46 @@ class SqlPermissionsRepository(
...
+    }
+
+    private fun getUserPermissions(id: String, unrestrictedUser: UserPermission?): Optional<UserPermission> {
+        // Check if the user exists and if they do are they an admin
+        val isAdmin = withRetry(RetryCategory.READ) {
+            jooq.select(USER.ADMIN)
+                .from(USER)
+                .where(USER.ID.eq(id))
+                .fetchOne(USER.ADMIN)
+        } ?: return Optional.empty()
```
mattgogerly, line 132: When would this be null? The withRetry call below doesn't have the null check.

mattgogerly (commented):
So if I'm understanding this right the idea is to only return the Roles a user has, rather than the entire UserPermission object that has accounts/applications etc?

fiat-roles/src/main/java/com/netflix/spinnaker/fiat/permissions/PermissionsRepository.java
deverton: Yeah, that could do with a rewording. It's basically 

> Finds all users with at least one of the set of specified roles and returns a map of users all their associated roles.

I'll fix that up.

fiat-sql/src/main/kotlin/com/netflix/spinnaker/fiat/permissions/SqlPermissionsRepository.kt
deverton: When the user doesn't exist in the Fiat permission repository. The logic here should match the RedisPermissionRepository.

deverton:
Yup, if you look at the role syncing process it actually completely recreates the `UserPermission` objects just based on the role data. The key bit is here https://github.com/spinnaker/fiat/blob/09668629f2cf7504955f762f99eb5098fd42b70d/fiat-roles/src/main/java/com/netflix/spinnaker/fiat/roles/UserRolesSyncer.java#L266 where you can see it just iterates over each user passing in the roles to the permission resolver.

For the SQL repository it's relatively expensive to read in the `UserPermission` object completely so just returning the `Role` instances is much quick. I assume it would be the same for the Redis repository but haven't tried to update that yet.

mattgogerly:
> Yup, if you look at the role syncing process it actually completely recreates the `UserPermission` objects just based on the role data. The key bit is here
> 
> https://github.com/spinnaker/fiat/blob/09668629f2cf7504955f762f99eb5098fd42b70d/fiat-roles/src/main/java/com/netflix/spinnaker/fiat/roles/UserRolesSyncer.java#L266
> 
> where you can see it just iterates over each user passing in the roles to the permission resolver.
> For the SQL repository it's relatively expensive to read in the `UserPermission` object completely so just returning the `Role` instances is much quick. I assume it would be the same for the Redis repository but haven't tried to update that yet.

Awesome - Fiat performance is brutal so any gains are very much welcome. Thanks for working on this!

deverton:
Had some time so added in an optimised implementation for Redis as well.

mattgogerly:
@deverton this still being worked on?

deverton:
@mattgogerly it's been running in production for the last two months and mostly fixed our issues with sync and auth times. It's still not perfect, but is a big help to performance.

mattgogerly (approved):

PR #901, opened 2022-02-01
perf(redis): Use SMEMBERS for redis repository scanSet

Nessex:
This was originally submitted here: https://github.com/spinnaker/fiat/pull/889
These new PRs are splitting the original PR into pieces for easier consumption.

## WHAT

This replaces the request per item done by the current scanSet with an SMEMBERS scan of the whole set in a single request stream.

This significantly improves roles sync time, redis CPU usage and redis network usage. For benchmark results, please see the [original PR](https://github.com/spinnaker/fiat/pull/889).

## WHY

In benchmarks, this provided a nearly 4x performance increase for the Redis permissions repository roles sync. Specifically, from a sync time of ~2 minutes down to ~30 seconds. We're still encountering issues with the duration of /roles/sync getting longer and longer as we add more users, more applications etc. Ideally, the sync would take less than 20 seconds to keep the sync time less than various timeouts of callers of Fiat.


mattgogerly (commented):
LGTM but not an expert on Fiat

fiat-roles/src/main/java/com/netflix/spinnaker/fiat/permissions/RedisPermissionsRepository.java
```diff
@@ -460,20 +460,14 @@ public void remove(@NonNull String id) {
...
-              });
-      results.addAll(
-          result.getResult().stream().map(SafeEncoder::encode).collect(Collectors.toList()));
-      cursor.set(result.getCursorAsBytes());
-    } while (!Arrays.equals(cursor.get(), ScanParams.SCAN_POINTER_START_BINARY));
-    return results;
+    return redisClientDelegate
+        .withBinaryClient(
+            jedis -> {
+              return jedis.smembers(key);
```
deverton, line 466: Isn't this going to cause Redis to block for the duration of the `SMEMBERS` call?

Nessex: Yes, almost all redis commands block IIRC, and this one is definitely included. However, even with tens of thousands of set members, this happens in <1msec in our case (and from prior experience this should be typical of redis). Compare that to the extreme overhead of having individual commands run for each set member, and it's a clear win overall.

This change also improves the consistency as the set can change from under redis between calls, e.g. this from the scan docs:

> * A given element may be returned multiple times. It is up to the application to handle the case of duplicated elements, for example only using the returned elements in order to perform operations that are safe when re-applied multiple times.
> * Elements that were not constantly present in the collection during a full iteration, may be returned or not: it is undefined.

Finally, given fiat itself locks certain operations while doing a roles sync, I don't think there is too much risk here. Fiat won't be responding to certain requests (and therefore, running other redis commands) anyway until the sync process is finished.

deverton (approved):

mattgogerly (approved):

mattgogerly:
@Mergifyio update

PR #902, opened 2022-02-01
perf(redis): RedisPermissionsRepository::getAllByRoles use parallelStream

Nessex:
This was originally submitted here: https://github.com/spinnaker/fiat/pull/889
These new PRs are splitting the original PR into pieces for easier consumption.

## WHAT

This makes getAllByRoles a mostly parallel operation by replacing the existing logic with a parallelStream.

## WHY

In benchmarks, this provided a nearly 2x performance increase for the Redis permissions repository roles sync. Specifically, from a sync time of ~2 minutes down to ~1 minute.

We're still encountering issues with the duration of /roles/sync getting longer and longer as we add more users, more applications etc. Ideally, the sync would take less than 20 seconds to keep the sync time less than various timeouts of callers of Fiat.


fiat-roles/src/main/java/com/netflix/spinnaker/fiat/permissions/RedisPermissionsRepository.java
```diff
@@ -402,11 +402,12 @@ public void putAllById(Map<String, UserPermission> permissions) {
...
     }
 
-    final Set<String> uniqueUsernames = new HashSet<>();
-    for (String role : new HashSet<>(anyRoles)) {
-      uniqueUsernames.addAll(
-          scanSet(roleKey(role)).stream().map(String::toLowerCase).collect(Collectors.toSet()));
-    }
+    final Set<String> uniqueUsernames =
+        new HashSet<>(anyRoles)
+            .parallelStream()
```
deverton, line 407: There were some concerns expressed here https://github.com/spinnaker/fiat/pull/830#issuecomment-777111793 that parallel calls to Redis might overwhelm it. Have you seen any of that in your testing?

The other thing is, it might be worth creating a dedicated thread pool for this rather than relying on the default fork-join pool. At least that way you can tweak the parallelism from config.

deverton (changes requested):

fiat-roles/src/main/java/com/netflix/spinnaker/fiat/permissions/RedisPermissionsRepository.java
Nessex: > There were some concerns expressed here https://github.com/spinnaker/fiat/pull/830#issuecomment-777111793 that parallel calls to Redis might overwhelm it. Have you seen any of that in your testing?

Not at all, I've seen redis do way more request throughput than our fiat instances can produce, and ours are pretty busy. The library already manages a connection pool and runs one request at a time in that connection pool, so I think we can trust the driver to do the right thing here. Since the last wave of improvements, there are far less expensive calls too (e.g. no more HM___ operations, just GET / SET etc.) so it should be even less of a concern.

> The other thing is, it might be worth creating a dedicated thread pool for this rather than relying on the default fork-join pool. At least that way you can tweak the parallelism from config.

Hmm, you're probably right here. I haven't spent any time working with parallel streams in Java, and pretty much just translated what I would do with rayon in rust. I'll take a deeper look at this over the next week.

Nessex: Added a thread pool for the sync process. That way, if I create PRs for the other parallelization tasks they can share a thread pool, but not hog the default one. This also lets you set the number of threads via the following `fiat-local.yml`:

```
fiat:
  redis:
    repository:
      syncThreads: 8
```

deverton (approved):

mattgogerly (approved):

mattgogerly:
@Mergifyio update

PR #960, opened 2022-06-22
fix(roles): Ensure account manager role is cached

jvz:
This updates the Redis and SQL permission caching to support the newly
introduced account manager meta-role.

* Ensures the `accountManager` flag is copied between live data and cache data as the cache data is what ends up determining Fiat policy evaluations in practice.


dbyron-sf (approved):

link108:
backporting this, since backports for https://github.com/spinnaker/fiat/pull/964 requires it

link108:
@mergifyio backport release-1.27.x release-1.28.x

PR #961, opened 2022-07-06
perf(serviceAccounts): allows syncing service accounts without resolving user roles

qunayu:
Hi team,

As prefaced on [Slack](https://spinnakerteam.slack.com/archives/C091CCWRJ/p1656558976174469), we have been running our own Spinnaker cluster for a while, and recently we've noticed that saving service accounts have been timing out from front50 to fiat. We discovered that this was because Fiat runs a sync across all users with the associated roles for the requested service account upon every save, which means for a relatively large user base, this operation results in well over 20s of runtime. 

We ran rudimentary investigation and found that this user role sync does not seem to be necessary for saving these service accounts, since scheduled syncs will straighten out any missing users recently added to this role on the next run. 

Following this train of thought, I'm raising this first PR on Fiat's side, which introduces a new endpoint that saves the service account, and adds it as a resource for all currently sync-ed users with these roles.

This change has been tested on my local dev environment, which seems to work well. It reduces the runtime of saving a service account to a sub-second operation. This should be quite a safe change to role out, since any impact will rely on consumes explicitly calling this new endpoint (which nobody is doing at the moment).

If everyone is happy with this change, the next PR we'll raise will be on Front50, which will call out to this new endpoint when saving service accounts. I'll be sure to introduce a config parameter so that this feature is opt-in only, and won't negatively impact users if something goes wrong.

Thanks!

### DIFF
--- a/fiat-roles/src/main/java/com/netflix/spinnaker/fiat/roles/UserRolesSyncer.java
+++ b/fiat-roles/src/main/java/com/netflix/spinnaker/fiat/roles/UserRolesSyncer.java
@@ -107,6 +107,10 @@ public class UserRolesSyncer {
     return syncStrategy.syncAndReturn(roles);
   }
 
+  public long syncServiceAccount(String serviceAccountId, List<String> roles) {
+    return syncStrategy.syncServiceAccount(serviceAccountId, roles);
+  }
+
   private static String metricName(String name) {
     return "fiat.userRoles." + name;
   }
--- a/fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatService.java
+++ b/fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatService.java
@@ -82,6 +82,18 @@ public interface FiatService {
   long sync(@Body List<String> roles);
 
   /**
+   * Use to update a service account. As opposed to `sync`, this will not trigger a full sync for
+   * user role membership.
+   *
+   * @param serviceAccountId Name of the service account.
+   * @param roles The roles allowed for this service account.
+   * @return The number of non-anonymous users synced.
+   */
+  @POST("/roles/sync/serviceAccount/{serviceAccountId}")
+  long syncServiceAccount(
+      @Path("serviceAccountId") String serviceAccountId, @Body List<String> roles);
+
+  /**
    * @param userId The user being logged in
    * @param ignored ignored.
    * @return ignored.
--- a/fiat-roles/src/main/java/com/netflix/spinnaker/fiat/permissions/DefaultPermissionsResolver.java
+++ b/fiat-roles/src/main/java/com/netflix/spinnaker/fiat/permissions/DefaultPermissionsResolver.java
@@ -203,7 +203,8 @@ public class DefaultPermissionsResolver implements PermissionsResolver {
     return userToRoles;
   }
 
-  private Map<String, UserPermission> resolveResources(
+  @Override
+  public Map<String, UserPermission> resolveResources(
       @NonNull Map<String, Collection<Role>> userToRoles) {
     return userToRoles.entrySet().stream()
         .map(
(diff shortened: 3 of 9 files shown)


j-sandy:
@qunayu as you mentioned, this PR is one part of the solution, is it possible to raise the second part of that solution in front50 repo and tracking them using a new spinnaker issue [here](https://github.com/spinnaker/spinnaker/issues) ? It will make tracking of the issue and its solution easy. And, if both the parts of the solution are available for review, it will help in verifying the real time scenario. 
Thanks !

qunayu:
Thanks @j-sandy , unfortunately the front50 bit depends on fiat-api being updated to include the new endpoint introduced, but I'll raise a draft PR (though builds will remain red) for that bit soon.

qunayu:
Hey @j-sandy , I've created a PR for the front50 component here. https://github.com/spinnaker/front50/pull/1139. Thanks.

fiat-roles/src/main/java/com/netflix/spinnaker/fiat/roles/UserRolesSyncer.java
```diff
@@ -201,6 +201,41 @@ public long syncAndReturn(List<String> roles) {
     }
   }
 
+  public long syncServiceAccounts(String serviceAccountId, List<String> roles) {
```
mattgogerly, line 204: If this is syncing one service account can we remove the plural from the method name? :)

mattgogerly:
From a purely code perspective I only have one minor nitpick. 

I'm not really familiar with how Fiat works though, so will defer to anybody who has more understanding.

fiat-roles/src/main/java/com/netflix/spinnaker/fiat/roles/UserRolesSyncer.java
```diff
@@ -201,6 +201,41 @@ public long syncAndReturn(List<String> roles) {
...
+    // So that fiat will pull fresh permissions
+    permissionsResolver.clearCache();
+
+    // Store this service account as a user with associated resources via `resolveAndMerge`.
+    List<Role> convertedRoles =
+        roles.stream()
+            .map(extRole -> new Role().setSource(Role.Source.EXTERNAL).setName(extRole))
```
jasonmcintosh, line 211: One thing I've not dug, though I THINK it's ok for Fiat.  Just making sure this is if this is how normal service accounts are handled.  E.g. there is some sync behavior on service accounts to try to avoid calling API's of remove systems (when using google/github for auth or ldap).

jasonmcintosh:
RIGHT NOW i'd lean towards approving this... E.g. this looks like it SHOULD work and improve processing being a targeted sync.  There's one question I had on the behavior of service accounts in the repo, and I'm not 100% that there isn't a transitive impact that's potentially missed.  BUT I think this should work and be a definite improvement!

fiat-roles/src/main/java/com/netflix/spinnaker/fiat/roles/UserRolesSyncer.java
qunayu: Absolutely, thanks.

fiat-roles/src/main/java/com/netflix/spinnaker/fiat/roles/UserRolesSyncer.java
qunayu: I think so - it seems to be done here as a part of the normal sync as well https://github.com/spinnaker/fiat/commit/906ecd7724441bc4b4b4aac349306fa3eded9359#diff-9cc5e46be2367bea393fc1b1449ef306c0c9f6be2802833a267b644cdb4c7b4cR172

qunayu:
Just resolved conflicts and it's ready for another review. If no major concerns, we'd love to merge this ASAP, thanks!

jasonmcintosh (approved):

dogonthehorizon (approved):

PR #962, opened 2022-07-21
perf(fiat-roles): Cache long-running roles' synchronization process

chriskotula:
Role synchronization is a time- and memory-consuming process. When we get a large number of concurrent requests to synchronize, we observe performance degradation and significant memory consumption. It was noticed in an environment with a large number of users and roles.

Prior to starting new synchronization we verify if the cache already has a running sync process that’s associated with a key, where the cache key is a set of roles to synchronize.
If there’s no cache entry, we start a new synchronization process as we’ve been doing so far.. When the cache contains the key and the associated process is still in progress, we reuse the process that’s represented as a Future<> instance.
Feature is backwards compatible - when no additional properties in configuration are provided it runs “old” logic. If `fiat.roles.sych-cache.enabled=true` is present in configuration, application uses cache for running processes.


fiat-roles/src/main/java/com/netflix/spinnaker/fiat/roles/CallableCache.java
```diff
@@ -0,0 +1,81 @@
+/*
+ * Copyright 2022 Netflix, Inc.
```
mattgogerly, line 2: s/Netflix/Atlassian?

fiat-roles/src/main/java/com/netflix/spinnaker/fiat/roles/CallableCache.java
```diff
@@ -0,0 +1,81 @@
...
+    } finally {
+      lock.unlock();
+    }
+  }
+
+  private boolean isPresent(Key key) {
+    return this.cache.containsKey(key);
+  }
+
+  private boolean isExecutionInProgress(Future<Result> callableFuture) {
```
ovidiupopa07, line 72: I think this method can be static

fiat-roles/src/main/java/com/netflix/spinnaker/fiat/roles/UserRolesSyncStrategy.java
```diff
@@ -0,0 +1,75 @@
...
+    private final CallableCache<List<String>, Long> callableCache;
+
+    public CachedSynchronizationStrategy(Synchronizer synchronizer) {
+      this.synchronizer = synchronizer;
+      this.callableCache = new CallableCache<>();
+    }
+
+    public CachedSynchronizationStrategy(
+        Synchronizer synchronizer, CallableCache<List<String>, Long> cache) {
+      this.synchronizer = synchronizer;
```
ovidiupopa07, line 51: Not sure if it would make sense to validate that the constructor parameters are not null? Just to make sure that whoever is instantiating this class is not passing a null param by mistake?

fiat-roles/src/main/java/com/netflix/spinnaker/fiat/roles/Synchronizer.java
```diff
@@ -0,0 +1,238 @@
...
+    // Ensure we're going to reload app and service account definitions
+    permissionsResolver.clearCache();
+
+    while (true) {
+      try {
+        Map<String, Set<Role>> combo = new HashMap<String, Set<Role>>();
+        // force a refresh of the unrestricted user in case the backing repository is empty:
+        combo.put(UnrestrictedResourceConfig.UNRESTRICTED_USERNAME, new HashSet<Role>());
+        Map<String, Set<Role>> temp;
+        if (!(temp = this.getUserPermissions(roles)).isEmpty()) {
```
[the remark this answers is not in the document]
chriskotula: Good catch. I didn't want to edit original code in any way so I just extracted some parts to new class and left them the way they were before. I do agree however that such small things could be improved. 
Thank you very much.

fiat-roles/src/main/java/com/netflix/spinnaker/fiat/roles/UserRolesSyncStrategy.java
chriskotula: @ovidiupopa07 I used lombok's annotation @NonNull. Is that ok?

ovidiupopa07: yep, sure 👍

ovidiupopa07 (approved):

fiat-roles/src/main/java/com/netflix/spinnaker/fiat/roles/CallableCache.java
chriskotula: Fixed! Thank you for pointing that out! 👀

mattgogerly (approved):

dogonthehorizon (approved):

jervi:
A bit late to the game, but this PR makes Fiat unable to start if `fiat.write-mode.enabled` is set to `false`. The `Synchronizer` class requires a `ResourceProvidersHealthIndicator` bean, but this bean has the following conditional set: `@ConditionalOnExpression("${fiat.write-mode.enabled:true}")`.

PR #966, opened 2022-07-28
feat(api): Add support for more Spring Security APIs

jvz:
This implements a few relevant Spring Security features and updates Fiat to use them in addition to the newly added AccessControlled interface in Kork.

- Introduces Spinnaker-specific GrantedAuthority values for admins, account managers, and anonymous users.
- Adds conversion from a user's set of roles to a set of granted authorities based on Spring Security conventions (`ROLE_foo` style authorities).
- Adds an implementation for FiatPermissionEvaluator::hasPermission for checking permissions on AccessControlled objects.
- Updates to Permissions/UserPermission to support GrantedAuthority conversions and checks.
- Adds spring-security-core as an API dependency to fiat-core as we now reference GrantedAuthority, Authentication, and some other core security interfaces.
- Updates Authorization parsing to support case-insensitive versions as Spring Security defines some common ones in lowercase like read/write/etc.

### DIFF
--- a/fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/AuthenticatedRequestAuthenticationConverter.java
+++ b/fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/AuthenticatedRequestAuthenticationConverter.java
@@ -1,0 +1,42 @@
+/*
+ * Copyright 2022 Apple, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.fiat.shared;
+
+import com.netflix.spinnaker.security.AuthenticatedRequest;
+import java.util.List;
+import javax.servlet.http.HttpServletRequest;
+import org.springframework.security.authentication.AnonymousAuthenticationToken;
+import org.springframework.security.core.Authentication;
+import org.springframework.security.core.authority.AuthorityUtils;
+import org.springframework.security.web.authentication.AuthenticationConverter;
+import org.springframework.security.web.authentication.preauth.PreAuthenticatedAuthenticationToken;
+
+public class AuthenticatedRequestAuthenticationConverter implements AuthenticationConverter {
+  @Override
+  public Authentication convert(HttpServletRequest request) {
+    return AuthenticatedRequest.getSpinnakerUser()
+        .map(
+            user ->
+                (Authentication) new PreAuthenticatedAuthenticationToken(user, "N/A", List.of()))
+        .orElseGet(
+            () ->
+                new AnonymousAuthenticationToken(
+                    "anonymous",
+                    "anonymous",
+                    AuthorityUtils.createAuthorityList("ROLE_ANONYMOUS")));
+  }
+}
--- a/fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/AuthenticatedRequestAuthenticationConverter.java
+++ b/fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/AuthenticatedRequestAuthenticationConverter.java
@@ -25,6 +25,12 @@ import org.springframework.security.core.authority.AuthorityUtils;
 import org.springframework.security.web.authentication.AuthenticationConverter;
 import org.springframework.security.web.authentication.preauth.PreAuthenticatedAuthenticationToken;
 
+/**
+ * Provides an Authentication for an HTTP request using the current {@link
+ * AuthenticatedRequest#getSpinnakerUser()}.
+ *
+ * @see AuthenticatedRequest
+ */
 public class AuthenticatedRequestAuthenticationConverter implements AuthenticationConverter {
   @Override
   public Authentication convert(HttpServletRequest request) {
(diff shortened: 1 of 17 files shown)


fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatAuthenticationFilter.java
```diff
@@ -16,49 +16,62 @@
...
   }
 
   @Override
   public void doFilter(ServletRequest request, ServletResponse response, FilterChain chain)
       throws IOException, ServletException {
+    Authentication auth;
     if (!fiatStatus.isEnabled()) {
-      chain.doFilter(request, response);
-      return;
+      List<GrantedAuthority> authorities =
```
dbyron-sf, line 50: I'm a little confused on why this is happening when fiat is disabled.

fiat-api/src/test/groovy/com/netflix/spinnaker/fiat/shared/AccessControlledResource.java
```diff
@@ -0,0 +1,36 @@
...
+import com.netflix.spinnaker.fiat.model.Authorization;
+import com.netflix.spinnaker.fiat.model.resources.Permissions;
+import com.netflix.spinnaker.security.AccessControlled;
+import org.springframework.security.core.Authentication;
+
+public class AccessControlledResource implements AccessControlled {
+  private final Permissions permissions;
+
+  public AccessControlledResource(Permissions permissions) {
+    this.permissions = permissions;
```
dbyron-sf, line 28:
```suggestion
    this.permissions = Objects.requireNonNull(permissions);
```
?

fiat-core/src/main/java/com/netflix/spinnaker/fiat/model/SpinnakerAuthorities.java
```diff
@@ -0,0 +1,39 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.fiat.model;
+
+import org.springframework.security.core.GrantedAuthority;
+import org.springframework.security.core.authority.SimpleGrantedAuthority;
+
+public class SpinnakerAuthorities {
```
dbyron-sf, line 22: javadoc please

fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatAuthenticationFilter.java
jvz: I was originally thinking that making an anonymous admin the security context when fiat was disabled, but that can be handled elsewhere.

fiat-api/src/test/groovy/com/netflix/spinnaker/fiat/shared/AccessControlledResource.java
jvz: I mean, it's test code, but ok.

fiat-core/src/main/java/com/netflix/spinnaker/fiat/model/SpinnakerAuthorities.java
jvz: Updated.

fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/AuthenticatedRequestAuthenticationConverter.java
```diff
@@ -0,0 +1,42 @@
...
+import com.netflix.spinnaker.security.AuthenticatedRequest;
+import java.util.List;
+import javax.servlet.http.HttpServletRequest;
+import org.springframework.security.authentication.AnonymousAuthenticationToken;
+import org.springframework.security.core.Authentication;
+import org.springframework.security.core.authority.AuthorityUtils;
+import org.springframework.security.web.authentication.AuthenticationConverter;
+import org.springframework.security.web.authentication.preauth.PreAuthenticatedAuthenticationToken;
+
+public class AuthenticatedRequestAuthenticationConverter implements AuthenticationConverter {
```
dbyron-sf, line 28: sorry to keep nitpicking...I need all the help I can get keeping the various bits and pieces here straight.  javadoc please.

jvz: I documented that on the bean that constructs it. Or do you want it in both places? Moved? I'm not sure where the most appropriate place would be to document those as the settings are sort of mutually incompatible being the same interface for a singleton bean and all.

fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatAuthenticationConverter.java
```diff
@@ -0,0 +1,47 @@
...
+import java.util.List;
+import javax.servlet.http.HttpServletRequest;
+import lombok.RequiredArgsConstructor;
+import org.springframework.security.authentication.AnonymousAuthenticationToken;
+import org.springframework.security.core.Authentication;
+import org.springframework.security.web.authentication.AuthenticationConverter;
+import org.springframework.security.web.authentication.preauth.PreAuthenticatedAuthenticationToken;
+
+@RequiredArgsConstructor
+public class FiatAuthenticationConverter implements AuthenticationConverter {
```
dbyron-sf, line 31: javadoc please

fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatPermissionEvaluator.java
```diff
@@ -148,6 +150,22 @@ private static RetryHandler buildRetryHandler(
   @Override
   public boolean hasPermission(
       Authentication authentication, Object resource, Object authorization) {
+    if (!fiatStatus.isEnabled()) {
```
dbyron-sf, line 153: Does it make sense to use the `services.fiat.granted-authorities.enabled` flag here too?

jvz: Hmm, maybe? Though when the flag is disabled, this method would end up returning false anyways since the authentication won't have any granted authorities. I can update this.

fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/AuthenticatedRequestAuthenticationConverter.java
dbyron-sf: I definitely like the comment on the bean method.  Maybe the same comment here makes sense as well (without "When enabled..")?  I would like to see something here that explains what the class is for.

dbyron-sf: Thank you!

dbyron-sf (approved):

PR #1003, opened 2022-12-12
refactor(sql): Optimize read queries for getUserPermission

jvz:
This fixes a couple issues in the existing implementation of getUserPermission. First, the account manager flag is queried at the same time as the admin flag (which is now fixed thanks to better understanding of the jOOQ API). Secondly, the functions to fetch resources for a given user were doing a natural join in Kotlin rather than offloading the request to the database to join.

### DIFF
--- a/fiat-sql/src/main/kotlin/com/netflix/spinnaker/fiat/permissions/SqlPermissionsRepository.kt
+++ b/fiat-sql/src/main/kotlin/com/netflix/spinnaker/fiat/permissions/SqlPermissionsRepository.kt
@@ -123,17 +123,11 @@ class SqlPermissionsRepository(
 
     private fun getUserPermissions(id: String, unrestrictedUser: UserPermission?): Optional<UserPermission> {
         // Check if the user exists and if they do are they an admin
-        val isAdmin = withRetry(RetryCategory.READ) {
-            jooq.select(USER.ADMIN)
+        val (isAdmin, isAccountManager) = withRetry(RetryCategory.READ) {
+            jooq.select(USER.ADMIN, USER.ACCOUNT_MANAGER)
                 .from(USER)
                 .where(USER.ID.eq(id))
-                .fetchOne(USER.ADMIN)
-        } ?: return Optional.empty()
-        val isAccountManager = withRetry(RetryCategory.READ) {
-            jooq.select(USER.ACCOUNT_MANAGER)
-                .from(USER)
-                .where(USER.ID.eq(id))
-                .fetchOne(USER.ACCOUNT_MANAGER)
+                .fetchOne()
         } ?: return Optional.empty()
 
         val userPermission = UserPermission()
@@ -141,12 +135,7 @@ class SqlPermissionsRepository(
             .setAdmin(isAdmin)
             .setAccountManager(isAccountManager)
 
-        val resourceIds = getUserPermissionsRecords(id)
-        val resourceRecords = fetchResourceRecords(resourceIds)
-
-        resourceIds.mapNotNull { resourceRecords[it] }.forEach {
-            userPermission.addResource(it)
-        }
+        addUserResources(userPermission)
 
         if (unrestrictedUser != null) {
             userPermission.merge(unrestrictedUser)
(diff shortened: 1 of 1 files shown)


fiat-sql/src/main/kotlin/com/netflix/spinnaker/fiat/permissions/SqlPermissionsRepository.kt
```diff
@@ -351,6 +312,23 @@ class SqlPermissionsRepository(
         }
     }
 
+    private fun addUserResources(userPermission: UserPermission) {
+        withRetry(RetryCategory.READ) {
+            jooq.selectFrom(RESOURCE.naturalJoin(PERMISSION))
```
jvz, line 317: I'll note that the bulk of the optimization is in this line. When the permission and resource tables were initially split, foreign keys were added for the resource name and resource type to link these tables together. However, this query wasn't ever updated to make use of this natural join relation. Read-side chunking of the query is no longer necessary as we can let the database plan out how to chunk things itself when forming the row result set.

dbyron-sf:
Thanks for this, it looks awesome!  Any chance of seeing `explain` output before and after the PR so we can see in more detail what's changed with the resulting queries?

jvz:
I didn't bother with an explain because this is SQL 101 level stuff.

jvz:
Let me see what data I can get.

xibz:
I like this change. This was doing multiple SQL calls before, and now is only doing a single SQL call which will improve network latency and DB calls.

jvz:
It will be tricky to do an explain on the before query because that was being split up into tons of tiny queries from both tables just to manually join the data, defeating the purpose of foreign keys and indices. Testing the updated query on my entire user database returned within a second.

dbyron-sf:
> It will be tricky to do an explain on the before query because that was being split up into tons of tiny queries from both tables just to manually join the data, defeating the purpose of foreign keys and indices. Testing the updated query on my entire user database returned within a second.

Any chance for an explain on the new query?

jvz:
Yes, though it'll take me a year to get it to you. Happy new year! 😂

dbyron-sf:
🤣

deverton:
I no longer work on Spinnaker, but I think when I added the chunking in it was necessary to cope with the number of rows we had in the `resource` table. We ran in to a lot of problems with Aurora MySQL not coping with the query across roughly 100 million rows. The idea for chunking was taken from how the queries work within Clouddriver.

However that was a while and several patches ago so maybe this is fine now? :)

jvz:
The old queries are like this:

```
explain select fiat_resource.resource_name, fiat_resource.body from fiat_resource where fiat_resource.resource_type = 'account' and fiat_resource.resource_name in ('name1', 'name2', 'name3');

Index Scan using fiat_resource_pkey on fiat_resource  (cost=0.41..17.29 rows=1 width=71)
   Index Cond: (((resource_type)::text = 'account'::text) AND ((resource_name)::text = ANY ('{name1,name2,name3}'::text[])))
```

That query is executed several times, and its results are used to do lookups in the `fiat_permission` table.

An explain of the updated query (or a variation of it; they all have similar query plans):

```
explain select body from fiat_resource r where exists ( select 1 from fiat_permission p where r.resource_name = p.resource_name and r.resource_type = p.resource_type and p.fiat_user_id = 'msicker@apple.com');

Hash Join  (cost=9447.95..11993.25 rows=9091 width=63)
   Hash Cond: (((r.resource_name)::text = (p.resource_name)::text) AND ((r.resource_type)::text = (p.resource_type)::text))
   ->  Seq Scan on fiat_resource r  (cost=0.00..2243.96 rows=57396 width=76)
   ->  Hash  (cost=9311.59..9311.59 rows=9091 width=17)
         ->  Index Only Scan using fiat_permission_pkey on fiat_permission p  (cost=0.56..9311.59 rows=9091 width=17)
               Index Cond: (fiat_user_id = 'msicker@apple.com'::text)
```

The exact plan chosen seems to depend on how Postgres (or whatever database you're using) chooses whether to use an index scan or a sequential scan. For example, running a similar explain query on an empty database:

```
explain select body from fiat_resource r where exists (select 1 from fiat_permission p where r.resource_name = p.resource_name and r.resource_type = p.resource_type and p.fiat_user_id = 'msicker@apple.com');

Nested Loop  (cost=0.28..16.58 rows=1 width=32)
  ->  Index Only Scan using fiat_permission_pkey on fiat_permission p  (cost=0.14..8.16 rows=1 width=1032)
        Index Cond: (fiat_user_id = 'msicker@apple.com'::text)
  ->  Index Scan using fiat_resource_pkey on fiat_resource r  (cost=0.14..8.16 rows=1 width=1064)
        Index Cond: (((resource_type)::text = (p.resource_type)::text) AND ((resource_name)::text = (p.resource_name)::text))
```

I'm looking in to making some larger simulations to test things out. It's possible that MySQL doesn't handle this query efficiently, or it's possible that I need to simulate a 100M+ record Fiat instance.

deverton:
Looking at some of the history of the changes, it looks like we were closer to 21 million permissions which doesn't seem so bad. We must have been having some other issues with our config that made MySQL unhappy.

I'd say go with the change you've got here since it does make far more sense to have the database do the join

jvz:
Been running this change in production for a while now. Would be nice to see this merged.

dbyron-sf:
@mergifyio update

PR #1007, opened 2023-01-20
fix(fiat-roles): Fiat fails to start if write mode is disabled

jervi:
Spring refuses to start due to a missing bean when write mode is disabled. The roles sync should not be enabled at all when write mode is disabled, so this fix disables both sync strategies and the Synchronizer class if read mode is disabled. Fixes bug introduced in https://github.com/spinnaker/fiat/pull/962.

Error message:

```
***************************
APPLICATION FAILED TO START
***************************
Description:
Parameter 0 of constructor in com.netflix.spinnaker.fiat.roles.Synchronizer required a bean of type 'com.netflix.spinnaker.fiat.config.ResourceProvidersHealthIndicato>
Action:
Consider defining a bean of type 'com.netflix.spinnaker.fiat.config.ResourceProvidersHealthIndicator' in your configuration.
```


jervi:
@spinnaker/oss-reviewers PTAL :)

mattgogerly (approved):

jasonmcintosh (approved):

PR #1012, opened 2023-02-06
fix: Should fix the deletion of permissions when resource name is uppercase

ovidiupopa07:
After the update to make all resource names lowercase, the user role sync process deletes the permission if the resource name (provided by the ResourceProvider) is in uppercase and the one in DB is in lowercase.


kkotula (approved):
👏🏻

jasonmcintosh (approved):

jasonmcintosh:
@Mergifyio backport release-1.28.x release-1.29.x release-1.27.x

PR #1021, opened 2023-02-17
chore(dependencies): remove residual dependency on groovy-all in fiat

j-sandy:


dbyron-sf (approved):

build.gradle
```diff
@@ -70,7 +70,6 @@ subprojects {
       testImplementation "org.hamcrest:hamcrest-core"
       testRuntimeOnly "cglib:cglib-nodep"
       testRuntimeOnly "org.objenesis:objenesis"
-      testImplementation "org.codehaus.groovy:groovy-all"
```
dbyron-sf, line 73: I verified that the same tests run both before and after this change.  There are some tests not running on master already which we'll address in a followup PR.

dbyron-sf (approved):

PR #1025, opened 2023-03-01
fix(tests): Introduce junit5 vintage engine for running junit4 test cases over junit5 in fiat

j-sandy:
Spring boot 2.4.x removed JUnit5 vintage engine from [spring-boot-starter-test](https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-2.4-Release-Notes#junit-5s-vintage-engine-removed-from-spring-boot-starter-test). It is required for executing junit4 based test cases in fiat. So, introducing junit-vintage-engine dependency in build.gradle, using testRuntimeOnly() as suggested in section 3.1 of https://junit.org/junit5/docs/5.6.2/user-guide/index.pdf

After applying this fix, coverage increased from 87 to 180 test case executions.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -43,6 +43,7 @@ subprojects {
       }
       minHeapSize = "512m"
       maxHeapSize = "512m"
+      useJUnitPlatform()
     }
     tasks.withType(Javadoc) {
       failOnError = false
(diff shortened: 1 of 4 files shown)


build.gradle
```diff
@@ -69,6 +69,7 @@ subprojects {
       testImplementation "org.springframework:spring-test"
       testImplementation "org.hamcrest:hamcrest-core"
       testRuntimeOnly "cglib:cglib-nodep"
+      testRuntimeOnly "org.junit.vintage:junit-vintage-engine"
```
dbyron-sf, line 72: Isn't the idea to add
```
test {
   useJunitPlatform()
}
```
also?

j-sandy: Updated `useJUnitPlaform()`  in build.gradle and removed this from sub modules.

dbyron-sf (approved):

PR #1029, opened 2023-03-14
fix(logs): Redacted secret data in logs.

DanielaS12:
Redacted secret data from log in `FiatAccessDeniedExceptionHandler`. 

### Example:

```
17:59:18.180 [Test worker] ERROR com.netflix.spinnaker.fiat.shared.FiatAccessDeniedExceptionHandler - Encountered exception while processing request : with headers={Authorization=bearer **REDACTED**, X-Frame-Options=**REDACTED**, Content-Type=text/html}
org.springframework.security.access.AccessDeniedException: Forbidden
```

### DIFF
--- a/fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatAccessDeniedExceptionHandler.java
+++ b/fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatAccessDeniedExceptionHandler.java
@@ -19,8 +19,6 @@ package com.netflix.spinnaker.fiat.shared;
 import com.netflix.spinnaker.kork.api.exceptions.AccessDeniedDetails;
 import com.netflix.spinnaker.kork.web.exceptions.ExceptionMessageDecorator;
 import java.io.IOException;
-import java.util.Enumeration;
-import java.util.HashMap;
 import java.util.Map;
 import java.util.StringJoiner;
 import javax.servlet.http.HttpServletRequest;
@@ -38,12 +36,9 @@ public class FiatAccessDeniedExceptionHandler {
   private final Logger log = LoggerFactory.getLogger(getClass());
 
   private final DefaultErrorAttributes defaultErrorAttributes = new DefaultErrorAttributes();
+  private final HeadersRedactor headersRedactor = new HeadersRedactor();
 
   private final ExceptionMessageDecorator exceptionMessageDecorator;
-
-  private static final String SECRET_DATA_HEADER_FORMAT = "X-";
-  private static final String AUTHORIZATION_HEADER = "Authorization";
-  private static final String SECRET_DATA_VALUE = "**REDACTED**";
 
   public FiatAccessDeniedExceptionHandler(ExceptionMessageDecorator exceptionMessageDecorator) {
     this.exceptionMessageDecorator = exceptionMessageDecorator;
@@ -55,7 +50,7 @@ public class FiatAccessDeniedExceptionHandler {
       throws IOException {
     storeException(request, response, e);
 
-    Map<String, String> headers = requestHeaders(request);
+    Map<String, String> headers = headersRedactor.getRedactedHeaders(request);
 
     log.error(
         "Encountered exception while processing request {}:{} with headers={}",
@@ -106,32 +101,6 @@ public class FiatAccessDeniedExceptionHandler {
 
     if (authorizationFailure.hasAuthorization()) {
       sj.add("- required authorization:").add(authorizationFailure.getAuthorization().toString());
-    }
-  }
-
-  private Map<String, String> requestHeaders(HttpServletRequest request) {
-    Map<String, String> headers = new HashMap<>();
-
-    if (request.getHeaderNames() != null) {
-      for (Enumeration<String> h = request.getHeaderNames(); h.hasMoreElements(); ) {
-        String headerName = h.nextElement();
-        String headerValue = getRedactedHeaderValue(headerName, request.getHeader(headerName));
-        headers.put(headerName, headerValue);
-      }
-    }
-
-    return headers;
-  }
-
-  private String getRedactedHeaderValue(String headerName, String headerValue) {
-    if (headerName.startsWith(SECRET_DATA_HEADER_FORMAT)) {
-      return SECRET_DATA_VALUE;
-    } else {
-      if (headerName.contains(AUTHORIZATION_HEADER)) {
-        return headerValue.trim().split("\n")[0].split(" ")[0] + " " + SECRET_DATA_VALUE;
-      } else {
-        return headerValue;
-      }
     }
   }
 
--- a/fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatAccessDeniedExceptionHandler.java
+++ b/fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatAccessDeniedExceptionHandler.java
@@ -41,6 +41,10 @@ public class FiatAccessDeniedExceptionHandler {
 
   private final ExceptionMessageDecorator exceptionMessageDecorator;
 
+  private static final String SECRET_DATA_HEADER_FORMAT = "X-";
+  private static final String AUTHORIZATION_HEADER = "Authorization";
+  private static final String SECRET_DATA_VALUE = "**REDACTED**";
+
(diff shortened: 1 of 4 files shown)


DanielaS12:
@jasonmcintosh please review changes

fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatAccessDeniedExceptionHandler.java
```diff
@@ -111,14 +115,26 @@ private Map<String, String> requestHeaders(HttpServletRequest request) {
...
-        String headerValue = request.getHeader(headerName);
+        String headerValue = getRedactedHeaderValue(headerName, request.getHeader(headerName));
         headers.put(headerName, headerValue);
       }
     }
 
     return headers;
   }
 
+  private String getRedactedHeaderValue(String headerName, String headerValue) {
```
kkotula, line 126: Looks good to me!

There's one thing you can consider though. I would move the logic you added to a new class that's only purpose would be to redact headers. We achieve a couple of benefits that way:
1. There's better separation of concerns,
2. It's easier to unit test your logic, because no mocks are required and you do not need to deal with the logic of FiatAccessDeniedExceptionHandler,


Take a look at snippet below:
```
private class HeadersRedactor {

        private static final String SECRET_DATA_HEADER_FORMAT = "X-";
        private static final String AUTHORIZATION_HEADER = "Authorization";
        private static final String SECRET_DATA_VALUE = "**REDACTED**";

        Map<String, String> redact(Map<String, String> originalHeaders) {
            var redactedHeaders = new HashMap<String, String>(originalHeaders.size());
            originalHeaders.forEach((key, value) -> {
                
                String newValue;
                if (key.startsWith(SECRET_DATA_HEADER_FORMAT)) {
                    newValue = SECRET_DATA_VALUE;
                } else {
                    if (key.contains(AUTHORIZATION_HEADER)) {
                        newValue = value.trim().split("\n")[0].split(" ")[0] + " " + SECRET_DATA_VALUE;
                    } else {
                        newValue = value;
                        redactedHeaders.put(key, value);
                    }
                }
                redactedHeaders.put(newValue, value);
            });
            return redactedHeaders;
        }

    }
 ```

Feel free to keep your implementation as it is! I just wanted to share how I would do it. :)

File: fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatAccessDeniedExceptionHandler.java [hunk 2]
```diff
@@ -111,14 +115,26 @@ private Map<String, String> requestHeaders(HttpServletRequest request) {
...
+    if (headerName.startsWith(SECRET_DATA_HEADER_FORMAT)) {
+      return SECRET_DATA_VALUE;
+    } else {
+      if (headerName.contains(AUTHORIZATION_HEADER)) {
+        return headerValue.trim().split("\n")[0].split(" ")[0] + " " + SECRET_DATA_VALUE;
```
kkotula, line 131: This is technically correct, but we can make it slightly easier to read.  You can extract new private method `removeCredentialsFromAuthorizationHeader` or something like that. It will be instantly obvious what this string transformation does.

kkotula (approved):
LGTM! 🚀

fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/HeadersRedactor.java
```diff
@@ -0,0 +1,58 @@
+/*
+ * Copyright 2023 Netflix, Inc.
```
ovidiupopa07, line 2: Let's change the copyright to Armory Inc

fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/HeadersRedactor.java
```diff
@@ -0,0 +1,58 @@
...
+      for (Enumeration<String> h = request.getHeaderNames(); h.hasMoreElements(); ) {
+        String headerName = h.nextElement();
+        String headerValue = getRedactedHeaderValue(headerName, request.getHeader(headerName));
+        headers.put(headerName, headerValue);
+      }
+    }
+    return headers;
+  }
+
+  private String getRedactedHeaderValue(String headerName, String headerValue) {
```
ovidiupopa07, line 43: can be static

fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatAccessDeniedExceptionHandler.java
```diff
@@ -38,6 +36,7 @@ public class FiatAccessDeniedExceptionHandler {
   private final Logger log = LoggerFactory.getLogger(getClass());
 
   private final DefaultErrorAttributes defaultErrorAttributes = new DefaultErrorAttributes();
+  private final HeadersRedactor headersRedactor = new HeadersRedactor();
```
mattgogerly, line 39: Don't think this is used now it's static?

fiat-api/src/test/java/com/netflix/spinnaker/fiat/shared/HeadersRedactorTest.java
```diff
@@ -0,0 +1,44 @@
...
+import org.junit.jupiter.api.Test;
+import org.springframework.mock.web.MockHttpServletRequest;
+
+class HeadersRedactorTest {
+
+  MockHttpServletRequest request = new MockHttpServletRequest();
+  HeadersRedactor unit = new HeadersRedactor();
+
+  @Test
+  public void redactHeadersTest() {
```
ovidiupopa07, line 31: It would be good to add a descriptive name for the test to know at a glance what it is testing instead of going through the test method and reading all the lines. 

@DisplayName or a meaningful method name

ovidiupopa07 (changes requested):

fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/HeadersRedactor.java
```diff
@@ -0,0 +1,58 @@
...
+      for (Enumeration<String> h = request.getHeaderNames(); h.hasMoreElements(); ) {
+        String headerName = h.nextElement();
+        String headerValue = getRedactedHeaderValue(headerName, request.getHeader(headerName));
+        headers.put(headerName, headerValue);
+      }
+    }
+    return headers;
+  }
+
+  private static String getRedactedHeaderValue(String headerName, String headerValue) {
```
jasonmcintosh, line 43: IF I'm following the logic here correctly, this means that X-* headers would have their values masked.  As a rule, anything with X- should be included and everything else masked/not included.  For example, X-Spinnaker-User is a KEY header to display as that identifies the user who made a request.  This ALSO simplifies the rather complicated "Look for the header... then look at the header starting again later on" logic.  

Instead I'd recommend something like:
"IF starts with X-, keep it and include the value.  Otherwise, don't even include it masked" 
as probably a safer setup.

jasonmcintosh (changes requested):

fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/HeadersRedactor.java
kkotula: @jasonmcintosh wouldn't it make sense to keep the non X- headers with masked values for debugging purposes? I am not saying we need to keep em, but their presence shouldn't hurt us and they can help when verifying some issues.

fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/HeadersRedactor.java
```diff
@@ -0,0 +1,50 @@
...
+    if (!headerName.startsWith(SECRET_DATA_HEADER_FORMAT)) {
```
jasonmcintosh, line 44: Minor... but the static no longer matches the intent.  Could be renamed "WHITELIST-HEADER-PREFIX" or some "NON_SECRET_DATA_HEADER_PREFIX" or some such.

DanielaS12:
@jasonmcintosh please merge and backport (I don't have access)

ovidiupopa07 (approved):
LGTM

PR #1034, opened 2023-03-21
feat(exceptions): Add SpinnakerRetrofitErrorHandler and replace RetrofitError catch blocks

abe21412:
This pull request registers `SpinnakerRetrofitErrorHandler` with each `Retrofit.RestAdapter` and replaces each `RetrofitError` catch block with a catch-block using `SpinnakerServerException` or the appropriate subclass. This is part of an effort to consume `SpinnakerRetrofitErrorHandler` in each Spinnaker microservice, as detailed in [issue #5473](https://github.com/spinnaker/spinnaker/issues/5473)

### DIFF
--- a/fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatPermissionEvaluator.java
+++ b/fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatPermissionEvaluator.java
@@ -26,6 +26,7 @@ import com.netflix.spinnaker.fiat.model.UserPermission;
 import com.netflix.spinnaker.fiat.model.resources.Account;
 import com.netflix.spinnaker.fiat.model.resources.Authorizable;
 import com.netflix.spinnaker.fiat.model.resources.ResourceType;
+import com.netflix.spinnaker.kork.retrofit.exceptions.SpinnakerHttpException;
 import com.netflix.spinnaker.kork.telemetry.caffeine.CaffeineStatsCounter;
 import com.netflix.spinnaker.kork.web.exceptions.NotFoundException;
 import com.netflix.spinnaker.security.AccessControlled;
@@ -187,8 +188,11 @@ public class FiatPermissionEvaluator implements PermissionEvaluator {
                       try {
                         fiatService.canCreate(username, resourceType, resource);
                         return true;
-                      } catch (NotFoundException e) {
-                        return false;
+                      } catch (SpinnakerHttpException e) {
+                        if(e.getResponseCode() == HttpStatus.NOT_FOUND.value()){
+                          return false;
+                        }
+                        throw e;
                       }
                     });
               })
--- a/fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatPermissionEvaluator.java
+++ b/fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatPermissionEvaluator.java
@@ -26,8 +26,6 @@ import com.netflix.spinnaker.fiat.model.UserPermission;
 import com.netflix.spinnaker.fiat.model.resources.Account;
 import com.netflix.spinnaker.fiat.model.resources.Authorizable;
 import com.netflix.spinnaker.fiat.model.resources.ResourceType;
-import com.netflix.spinnaker.kork.retrofit.exceptions.SpinnakerHttpException;
-import com.netflix.spinnaker.kork.retrofit.exceptions.SpinnakerServerException;
 import com.netflix.spinnaker.kork.telemetry.caffeine.CaffeineStatsCounter;
 import com.netflix.spinnaker.kork.web.exceptions.NotFoundException;
 import com.netflix.spinnaker.security.AccessControlled;
@@ -191,16 +189,6 @@ public class FiatPermissionEvaluator implements PermissionEvaluator {
                         return true;
                       } catch (NotFoundException e) {
                         return false;
-                      } catch (SpinnakerHttpException e) {
-                        boolean shouldRetry = true;
-                        if (e.getResponseCode() == HttpStatus.BAD_REQUEST.value()) {
-                          shouldRetry = false;
-                        }
-                        e.setRetryable(shouldRetry);
-                        throw e;
-                      } catch (SpinnakerServerException e) {
-                        e.setRetryable(true);
-                        throw e;
                       }
                     });
               })
--- a/fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatPermissionEvaluator.java
+++ b/fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatPermissionEvaluator.java
@@ -26,7 +26,6 @@ import com.netflix.spinnaker.fiat.model.UserPermission;
 import com.netflix.spinnaker.fiat.model.resources.Account;
 import com.netflix.spinnaker.fiat.model.resources.Authorizable;
 import com.netflix.spinnaker.fiat.model.resources.ResourceType;
-import com.netflix.spinnaker.kork.exceptions.IntegrationException;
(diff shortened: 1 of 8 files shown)


dbyron-sf:
Thanks for this @abe21412 .  Would you mind writing out explicitly if there are any changes in behavior here.

fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatPermissionEvaluator.java
```diff
@@ -188,19 +190,20 @@ public boolean canCreate(String resourceType, Object resource) {
...
-                            case NOT_FOUND:
-                              return false;
-                            case BAD_REQUEST:
-                              shouldRetry = false;
-                          }
+                        if (e.getResponseCode() == HttpStatus.BAD_REQUEST.value()) {
+                          shouldRetry = false;
                         }
-                        IntegrationException ie = new IntegrationException(re);
+                        IntegrationException ie = new IntegrationException(e);
```
dbyron-sf, line 200: This is potentially a candidate for a followup PR, but I wonder if this exception wrapping is necessary anymore.  Like, we've got a SpinnakerHttpException here, which is already a child of SpinnakerException, and so already has its own retryable attribute.  Plus also, I'd love to depend on SpinnakerRetrofitErrorHandler to set that retryable attribute appropriately.

I don't see any code catching IntegrationException explicitly (except some kork-plugins tests, e.g. [here](https://github.com/spinnaker/kork/blob/b6fa1103b7b3b12734af39cb6f6ca2b04cdd59d6/kork-plugins/src/test/kotlin/com/netflix/spinnaker/kork/plugins/DependencyInjectionTest.kt#L87)), so from that perspective I think the change is OK.

abe21412: Ah, I didn't realize `SpinnakerException` had its own `setRetryable()` method. I can go ahead and use that instead.

dbyron-sf:
> > Thanks for this @abe21412 . Would you mind writing out explicitly if there are any changes in behavior here.
> 
> The only change in behavior is that using `SpinnakerRetrofitErrorHandler` makes errors from other spinnaker services more visible to users. Other than that, behavior should be the same as the changes here only refactor existing code to use more specific exception classes.

Excellent.  You don't happen to have any screenshots of specific places where users see better error messages?  Or specific scenarios where they would expect to?

abe21412:
> > > Thanks for this @abe21412 . Would you mind writing out explicitly if there are any changes in behavior here.
> > 
> > 
> > The only change in behavior is that using `SpinnakerRetrofitErrorHandler` makes errors from other spinnaker services more visible to users. Other than that, behavior should be the same as the changes here only refactor existing code to use more specific exception classes.
> 
> Excellent. You don't happen to have any screenshots of specific places where users see better error messages? Or specific scenarios where they would expect to?

I don't have any screenshots on hand but a scenario where we could expect to see better error messages could be if fiat queries front50 for service accounts and encounters an error from front50.

fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatPermissionEvaluator.java
```diff
@@ -188,19 +189,18 @@ public boolean canCreate(String resourceType, Object resource) {
...
-                          }
+                        if (e.getResponseCode() == HttpStatus.BAD_REQUEST.value()) {
+                          shouldRetry = false;
                         }
-                        IntegrationException ie = new IntegrationException(re);
-                        ie.setRetryable(shouldRetry);
-                        throw ie;
+                        e.setRetryable(shouldRetry);
+                        throw e;
+                      } catch (SpinnakerServerException e) {
```
dbyron-sf, line 201: Plus also, [SpinnakerServerException](https://github.com/spinnaker/kork/blob/b6fa1103b7b3b12734af39cb6f6ca2b04cdd59d6/kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/exceptions/SpinnakerServerException.java) doesn't mention retryable, so from [SpinnakerException](https://github.com/spinnaker/kork/blob/b6fa1103b7b3b12734af39cb6f6ca2b04cdd59d6/kork-exceptions/src/main/java/com/netflix/spinnaker/kork/exceptions/SpinnakerException.java), retryable is null...and then RetrySupport [retries](https://github.com/spinnaker/kork/blob/b6fa1103b7b3b12734af39cb6f6ca2b04cdd59d6/kork-exceptions/src/main/java/com/netflix/spinnaker/kork/core/RetrySupport.java#L38-L40).

fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatPermissionEvaluator.java
```diff
@@ -188,19 +189,18 @@ public boolean canCreate(String resourceType, Object resource) {
                       try {
                         fiatService.canCreate(username, resourceType, resource);
                         return true;
-                      } catch (RetrofitError re) {
+                      } catch (NotFoundException e) {
+                        return false;
+                      } catch (SpinnakerHttpException e) {
```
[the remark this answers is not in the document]
abe21412: We can omit setting to false but I think we still need to explicitly set retryable to true if the error code is not 400 or 404, right? [Because retryable is null by default](https://github.com/spinnaker/kork/blob/b6fa1103b7b3b12734af39cb6f6ca2b04cdd59d6/kork-exceptions/src/main/java/com/netflix/spinnaker/kork/exceptions/SpinnakerException.java#L45).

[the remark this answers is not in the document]
abe21412: oh, I just saw your other comment, so the behavior is the same whether retryable is set to null or true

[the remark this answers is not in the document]
dbyron-sf: Right!

fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatPermissionEvaluator.java
```diff
@@ -188,19 +187,8 @@ public boolean canCreate(String resourceType, Object resource) {
...
-                            case NOT_FOUND:
-                              return false;
-                            case BAD_REQUEST:
-                              shouldRetry = false;
-                          }
-                        }
-                        IntegrationException ie = new IntegrationException(re);
-                        ie.setRetryable(shouldRetry);
-                        throw ie;
+                      } catch (NotFoundException e) {
```
dbyron-sf, line 190: OK, this bit is finally sinking in.  Sorry if I pointed you in the wrong direction before.

[SpinnakerRetrofitErrorHandler](https://github.com/spinnaker/kork/blob/b6fa1103b7b3b12734af39cb6f6ca2b04cdd59d6/kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/exceptions/SpinnakerRetrofitErrorHandler.java) doesn't throw NotFoundException, so I don't think it helps to catch it here.  To get the same behavior as before I think we need to catch SpinnakerHttpException and check for HttpStatus.NOT_FOUND to decide whether to return false.

dbyron-sf (approved):

dbyron-sf:
@abe21412 looks like some spotless checks are failing.

dbyron-sf:
> @abe21412 looks like some spotless checks are failing.

I went ahead and pushed the cleanup.

dbyron-sf:
@mergifyio update

dbyron-sf:
I also took the liberty of removing the `com.squareup.retrofit:retrofit` dependency in fiat-api since it's no longer used.

PR #1060, opened 2023-06-07
feat(fiat): Suppress application details when updating permissions

kirangodishala:
- This PR allows skipping Application.details while caching the UserPermission.
- This is an opt-in feature which can be enabled by the following configuration:
```yaml
resource:
  provider:
    application:
      suppressDetails: true
```
- Fixes the issue:  https://github.com/spinnaker/spinnaker/issues/6842

### DIFF
--- a/fiat-roles/src/main/java/com/netflix/spinnaker/fiat/config/ResourceProviderConfig.java
+++ b/fiat-roles/src/main/java/com/netflix/spinnaker/fiat/config/ResourceProviderConfig.java
@@ -16,6 +16,8 @@
 
 package com.netflix.spinnaker.fiat.config;
 
+import com.google.common.collect.Sets;
+import java.util.Set;
 import lombok.Data;
 import org.springframework.boot.context.properties.ConfigurationProperties;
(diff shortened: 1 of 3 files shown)


mattgogerly:
Could you clarify what you mean by application.details? Do you have an example of a cache entry before and after this change?

kirangodishala:
> Could you clarify what you mean by application.details? Do you have an example of a cache entry before and after this change?

When fiat fetches applications from Front50 these details come along. 
image

kirangodishala:
> This seems like an interesting idea, though I wonder why we even cache things besides permissions in that table in the first place?

As I mentioned above, Application data primarily comes from Front50 (/v2/applications?restricted=false) and gets parsed into [Application](https://github.com/spinnaker/fiat/blob/master/fiat-core/src/main/java/com/netflix/spinnaker/fiat/model/resources/Application.java) object and the additional details were never removed.

mattgogerly (approved):

dbyron-sf (approved):

PR #1077, opened 2023-07-14
fix(roles-sync): fix CallableCache's NPE exception for caching synchronization strategy

kirangodishala:
This PR addresses: https://github.com/spinnaker/spinnaker/issues/6856

Made sure the calls to CallableCache are null-free to prevent NullPointerExceptions.


dbyron-sf:
Looks like the PR checks are failing because of linting that
```
$ ./gradlew :fiat-roles:spotlessApply
```
would fix.  Could use a test for this too.

dbyron-sf:
Looks like this is a fix to code added in https://github.com/spinnaker/fiat/pull/962 which first appeared in spinnaker 1.29.  That matches what https://github.com/spinnaker/spinnaker/issues/6856 says so backporting to release-1.29.x onwards.

dbyron-sf:
@mergifyio backport release-1.29.x release-1.30.x release-1.31.x

PR #1082, opened 2023-08-10
fix(ssl): Removed unused deprecated okHttpClientConfig from retrofitConfig.

DanielaS12:
This change removes unused deprecated class. Its instantiation caused fiat to fail when configuring SSL. Other then fixing an issue, the class was never used.

Tested on local environment on fiat 1.29.x, master and 1.30.x.

Before this change:
* fiat was throwing FileNotFound because OkHttpClientConfigurationProperties were not decrypted
  * the properties were not decrypted and the key-store could not be downloaded into a file
* this happened because  okHttpClientConfig instantiated OkHttpClientConfigurationProperties before the SecretBeanPostProcessor was instantiated

After the change:
* SecretBeanPostProcessor is instantiated first
* then it is used when OkHttpClientConfigurationProperties is instantiated  to decrypt each property

### DIFF
--- a/fiat-web/src/main/java/com/netflix/spinnaker/fiat/config/RetrofitConfig.java
+++ b/fiat-web/src/main/java/com/netflix/spinnaker/fiat/config/RetrofitConfig.java
@@ -21,7 +21,6 @@ import com.fasterxml.jackson.databind.DeserializationFeature;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.SerializationFeature;
 import com.google.common.collect.ImmutableList;
-import com.netflix.spinnaker.config.OkHttpClientConfiguration;
 import java.io.IOException;
 import java.util.List;
 import lombok.AllArgsConstructor;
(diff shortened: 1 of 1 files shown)


DanielaS12:
@jonsie @cfieber Could you please review this changes? Thank you!

dbyron-sf:
What config flags made fiat fail?

DanielaS12:
@dbyron-sf there were no config changes. The failure appeared after bumping up the kork version https://github.com/spinnaker/fiat/commit/b655872e5a9cf88573a4d6561733b5f70583c3aa, specifically this commit in kork https://github.com/spinnaker/kork/commit/a68c7f5dcf7ce1f8972a3d8de6910639d11c9da7

This change is paired with multiple commits in fiat, that omitted to clean the resource from RetrofitConfig.

kkotula (approved):
LGTM 🚀

DanielaS12:
@dbyron-sf yes it does not mention the configuration, but since it has multiple beans regarding secrets and secretManager,  my verified guess is that messes with the bean creation order.

DanielaS12:
@jonsie @cfieber Could you please review this changes? Thank you!

ovidiupopa07 (approved):

DanielaS12:
@dbyron-sf Can this be merged?

ovidiupopa07:
It looks good to me. Considering that with this PR we do some cleanup (delete unused field) that is messing with the SSL configuration.
I am going ahead and will merge this considering that all tests are passing.

jasonmcintosh:
@Mergifyio backport release-1.29.x release-1.30.x release-1.31.x

jasonmcintosh:
The kork PR broke this b/c it changed the ordering of bean loading/dependency handling :(  The lazy loading IN THEORY might have fixed this, but ... that's iffy.

PR #1096, opened 2023-08-31
Config simplifications

jvz:
Provided in multiple commits, I've modernized and simplified several config classes by reusing Kork and Spring Boot things.

### DIFF
--- a/fiat-web/src/main/java/com/netflix/spinnaker/fiat/config/FiatConfig.java
+++ b/fiat-web/src/main/java/com/netflix/spinnaker/fiat/config/FiatConfig.java
@@ -1,6 +1,5 @@
 package com.netflix.spinnaker.fiat.config;
 
-import com.google.common.collect.ImmutableList;
 import com.netflix.spectator.api.Registry;
 import com.netflix.spinnaker.config.PluginsAutoConfiguration;
 import com.netflix.spinnaker.fiat.model.Authorization;
@@ -24,7 +23,6 @@ import java.util.Collection;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import lombok.val;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
@@ -40,19 +38,19 @@ import org.springframework.scheduling.TaskScheduler;
 import org.springframework.scheduling.concurrent.ThreadPoolTaskScheduler;
 import org.springframework.web.servlet.config.annotation.ContentNegotiationConfigurer;
 import org.springframework.web.servlet.config.annotation.InterceptorRegistry;
-import org.springframework.web.servlet.config.annotation.WebMvcConfigurerAdapter;
+import org.springframework.web.servlet.config.annotation.WebMvcConfigurer;
 
 @Configuration
 @Import({RetrofitConfig.class, PluginsAutoConfiguration.class})
 @EnableConfigurationProperties(FiatServerConfigurationProperties.class)
-public class FiatConfig extends WebMvcConfigurerAdapter {
+public class FiatConfig implements WebMvcConfigurer {
 
   @Autowired private Registry registry;
 
   @Override
   public void addInterceptors(InterceptorRegistry registry) {
-    var pathVarsToTag = ImmutableList.of("accountName", "applicationName", "resourceName");
-    List<String> exclude = ImmutableList.of("BasicErrorController");
+    var pathVarsToTag = List.of("accountName", "applicationName", "resourceName");
+    List<String> exclude = List.of("BasicErrorController");
     MetricsInterceptor interceptor =
         new MetricsInterceptor(this.registry, "controller.invocations", pathVarsToTag, exclude);
     registry.addInterceptor(interceptor);
@@ -60,8 +58,7 @@ public class FiatConfig extends WebMvcConfigurerAdapter {
 
   @Override
   public void configureContentNegotiation(ContentNegotiationConfigurer configurer) {
-    super.configureContentNegotiation(configurer);
-    configurer.favorPathExtension(false).defaultContentType(MediaType.APPLICATION_JSON);
+    configurer.defaultContentType(MediaType.APPLICATION_JSON);
   }
 
   @Bean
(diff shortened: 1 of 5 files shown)


fiat-web/src/main/java/com/netflix/spinnaker/fiat/config/RetrofitConfig.java
```diff
@@ -33,37 +30,19 @@
...
-
-  @Value("${ok-http-client.retries.max-elapsed-backoff-ms:5000}")
-  @Setter
-  private long maxElapsedBackoffMs;
-
   @Bean
   @Primary
-  ObjectMapper objectMapper() {
-    return new ObjectMapper()
-        .configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false)
```
jasonmcintosh, line 64: Intentionally re-enabling this?  AKA removing the disable would... then trigger failures on unknown properties?

fiat-web/src/main/java/com/netflix/spinnaker/fiat/config/RetrofitConfig.java
```diff
@@ -33,37 +30,19 @@
...
 import retrofit.RestAdapter;
 
-/** This package is placed in fiat-core in order to be shared by fiat-web and fiat-shared. */
+/** Provides base beans used for configuring Retrofit REST client facades. */
 @Configuration
 public class RetrofitConfig {
 
-  @Value("${ok-http-client.connection-pool.max-idle-connections:5}")
-  @Setter
-  private int maxIdleConnections;
```
jasonmcintosh, line 46: THANK YOU :) Switching to kork based stuff.... finally!

fiat-web/src/main/java/com/netflix/spinnaker/fiat/config/RetrofitConfig.java
jasonmcintosh: Per slack, the Jackson2ObjectMapperBuilder sets this by default, so by switching to it we're a touch cleaner.

jasonmcintosh (approved):

fiat-web/src/main/java/com/netflix/spinnaker/fiat/config/FiatConfig.java
```diff
@@ -123,11 +120,11 @@ public DefaultServiceAccountPredicateProvider defaultServiceAccountPredicateProv
...
   /**
    * This AuthenticatedRequestFilter pulls the email and accounts out of the Spring security context
-   * in order to enabling forwarding them to downstream components.
+   * in order to enable forwarding them to downstream components.
    */
   @Bean
-  FilterRegistrationBean authenticatedRequestFilter() {
-    val frb = new FilterRegistrationBean(new AuthenticatedRequestFilter(true));
+  FilterRegistrationBean<AuthenticatedRequestFilter> authenticatedRequestFilter() {
+    var frb = new FilterRegistrationBean<>(new AuthenticatedRequestFilter(true));
```
jasonmcintosh, line 127: MINOR thing... why the var vs. val?

fiat-web/src/main/java/com/netflix/spinnaker/fiat/config/ResourcesConfig.java
```diff
@@ -35,45 +30,18 @@
...
-            new Ok3Client(
-                clientProvider.getClient(new DefaultServiceEndpoint("front50", front50Endpoint))))
-        .setConverter(new JacksonConverter(objectMapper))
-        .setErrorHandler(SpinnakerRetrofitErrorHandler.getInstance())
-        .setLogLevel(retrofitLogLevel)
-        .setLog(new Slf4jRetrofitLogger(Front50Api.class))
-        .build()
-        .create(Front50Api.class);
+  Front50Api front50Api(@Value("${services.front50.base-url}") String front50Endpoint) {
+    return provider.getService(
```
dbyron-sf, line 43: I believe this ends up with [RetrofitServiceFactory](https://github.com/spinnaker/kork/blob/825c81a177d5c3b11a286d664759ea2786a4bd87/kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/RetrofitServiceFactory.java#L51) creating the returned object.  At the moment that code doesn't use
```
.setErrorHandler(SpinnakerRetrofitErrorHandler.getInstance())
```
so this ends up being a change in behavior.  Adding the setErrorHandler call to RetrofitServiceFactory is in the plans, but is gonna take some care (and potentially a config flag controlling whether it happens or not) since I believe both gate and echo use it, and we need to coordinate changes in handling of RetrofitError to Spinnaker*Exception.

jvz: Isn't the new error handler for running in Retrofit2? Right now, `ServiceClientFactory` implementations check for the use of the `Call<T>` API for determining which Retrofit version to use.

fiat-web/src/main/java/com/netflix/spinnaker/fiat/config/FiatConfig.java
jvz: `val` is a Lombok extension while `var` is part of Java since 11 or so.

fiat-web/src/main/java/com/netflix/spinnaker/fiat/config/ResourcesConfig.java
dbyron-sf: SpinnakerRetrofitErrorHandler is for retrofit1.

jvz: But Retrofit only throws `RetrofitError` in the first place?

dbyron-sf: Yup, retrofit1 throws RetrofitError, and SpinnakerRetrofitErrorHandler turns RetrofitError into SpinnakerServerException (and children).

jvz: Oh, so the point of the error handler Bean is to translate the retrofit exception into a spinnaker one?

dbyron-sf: It's not a bean, but yes.

jvz:
No longer relevant; will be included in a PR to spinnaker/spinnaker.

PR #1131, opened 2024-01-03
Updates for #6911 and #6918

jvz:
Contains two small commits separated by the two relevant issues:

* https://github.com/spinnaker/spinnaker/issues/6918 (small change which required similar dependency updates)
* https://github.com/spinnaker/spinnaker/issues/6911 (linked with changes published in Kork)

### DIFF
--- a/fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/AuthenticatedRequestAuthenticationConverter.java
+++ b/fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/AuthenticatedRequestAuthenticationConverter.java
@@ -17,11 +17,8 @@
 package com.netflix.spinnaker.fiat.shared;
 
 import com.netflix.spinnaker.security.AuthenticatedRequest;
-import com.netflix.spinnaker.security.SpinnakerAuthorities;
-import com.netflix.spinnaker.security.SpinnakerUsers;
 import java.util.List;
 import javax.servlet.http.HttpServletRequest;
-import org.springframework.security.authentication.AnonymousAuthenticationToken;
 import org.springframework.security.core.Authentication;
 import org.springframework.security.web.authentication.AuthenticationConverter;
 import org.springframework.security.web.authentication.preauth.PreAuthenticatedAuthenticationToken;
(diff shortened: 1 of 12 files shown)


fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/AuthenticatedRequestAuthenticationConverter.java
```diff
@@ -41,8 +42,8 @@ public Authentication convert(HttpServletRequest request) {
         .orElseGet(
             () ->
                 new AnonymousAuthenticationToken(
-                    "anonymous",
-                    "anonymous",
-                    AuthorityUtils.createAuthorityList("ROLE_ANONYMOUS")));
+                    SpinnakerUsers.ANONYMOUS,
```
dbyron-sf, line 45: notes for reviewers -- the new code looks like it does exactly the same thing as the old code, so this is "just" a refactor.

fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatAuthenticationConfig.java
```diff
@@ -143,6 +144,8 @@ protected void configure(HttpSecurity http) throws Exception {
           .exceptionHandling()
           .and()
           .anonymous()
+          // match the same anonymous userid as expected elsewhere
```
dbyron-sf, line 147: How is this different from `.anonymous()` above?  Can you write a test that demonstrates this?

jvz: The default settings in this DSL set the principal to the string `anonymousUser`. The DSL is tricky to test since it's mostly a DSL for setting up beans.

jvz: So using the updated Spring Security DSL, this whole method would look more like the following:

```java
      http.servletApi(Customizer.withDefaults())
          .exceptionHandling(Customizer.withDefaults())
          .anonymous(anonymous -> anonymous.principal(SpinnakerUsers.ANONYMOUS))
          .addFilterBefore(
              new FiatAuthenticationFilter(fiatStatus, authenticationConverter),
              AnonymousAuthenticationFilter.class);
```

dbyron-sf: I didn't see before. `.principal(SpinnakerUsers.ANONYMOUS)` modifies how `.anonymous()` works...it's not another mechanism in addition to it.

jvz: I'll figure out some sort of test.

jvz: Turns out testing this is futile. I've extracted some constants to make this more consistent.

jvz:
No longer relevant; will be included in a PR to spinnaker/spinnaker.

PR #1155, opened 2024-04-26
refactor(api): change FiatPermissionEvaluator to implement UserPermis…

dbyron-sf:
…sionEvaluator

instead of PermissionEvaluator, and mark
```
public boolean hasPermission(
      String username, Serializable resourceName, String resourceType, Object authorization)
```
as `@Override`.

This makes this method available to e.g. S3ArtifactStoreGetter so it can authenticate by user.  In some pipeline execution scenarios in orca (e.g. using #fetchReference in an Evaluate Variables stage), this is necessary since SecurityContextHolder.getContext() is null.

depends on https://github.com/spinnaker/kork/pull/1179.


xibz (approved):

dbyron-sf:
@mergifyio update

PR #1196, opened 2024-12-26
chore(dependency): upgrade spring boot from 2.7.x to 3.0.x and spring cloud from 2021.0.x to 2022.0.x

j-sandy:
Upgrading spring boot 2.7.18 to 3.0.13 and spring cloud 2021.0.8 to 2022.0.5. Spring cloud release 2022.0.x is compatible with spring boot 3.0.x. https://github.com/spring-cloud/spring-cloud-release/wiki/Supported-Versions#supported-releases ====================================================== 
refactor(dependency): replace javax with jakarta during upgrade of spring boot 3.0.x

Spring Boot 3.0 has migrated from Java EE to Jakarta EE APIs for all dependencies. https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-3.0-Release-Notes#third-party-library-upgrades So, replacing the imports with `javax` to `jakarta`.


fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatAuthenticationConfig.java
```diff
@@ -92,10 +92,24 @@ AuthenticationConverter defaultAuthenticationConverter() {
     return new AuthenticatedRequestAuthenticationConverter();
   }
 
+  // Removing bean of WebSecurityConfigurerAdapter, as in spring security 6.0 it is no more
```
dbyron-sf, line 95: Wouldn't this be better in the commit message?

j-sandy: Updated.

PR #1206, opened 2025-03-22
fix(retrofit2): fix multiple slash issue in baseUrl of retrofit2 client

kirangodishala:
- When a retrofit client baseUrl has multiple slashes(ex: `https://somehost/api/v3/`), the resulting url changes based on the whether there is a leading slash in the endpoint or not. If the endpoint is prefixed with a `/` (ex: `/orgs/{org}/members`) the resulting url omits the segment post first `/` from baseUrl.  

- The example above results in `https://somehost/orgs/someorg/members` instead of `https://somehost/api/v3/orgs/someorg/members`. 

- This PR addresses the issue by removing the leading `/` from all the endpoints and making sure the baseUrl ends with a trailing `/`

- A test is added to demonstrate the issue.

### DIFF
--- a/fiat-core/src/main/java/com/netflix/spinnaker/fiat/util/RetrofitUtils.java
+++ b/fiat-core/src/main/java/com/netflix/spinnaker/fiat/util/RetrofitUtils.java
@@ -1,0 +1,42 @@
+/*
+ * Copyright 2025 OpsMx, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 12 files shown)


fiat-core/src/main/java/com/netflix/spinnaker/fiat/util/RetrofitUtils.java
```diff
@@ -0,0 +1,42 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.fiat.util;
+
+import okhttp3.HttpUrl;
+
+public class RetrofitUtils {
```
dbyron-sf, line 21: Seems like this belongs in kork.  We've got copies of this sprinkled around a few repos, right?  Happy to get these fixes in for now, but let's plan to centralize.

dbyron-sf (approved):

dbyron-sf:
@mergifyio backport release-1.37.x

PR #1213, opened 2025-05-04
feat(fiat/google-groups): expand indirect google groups for emails

himanhsugusain:
Google groups list api does not return the groups that the user is indirectly part of.
This change recursively traverse each group and get its parent group, we also keep track of already visited groups and not query it again.
Google groups does not allow one group to be in subgroup of itself, so that case check is not needed.

Eg.
User A if member of Group G1 and G1 is member of G2, so A is member of G2 indireclty.
Before:
Access provided to group G2 will not be inherited by A.
After
Access provided to G2 will be inherited by A.

PS: This is how Google own permission model work. Too bad they don't provide a direct api for it.

This also updates `com.google.api.client.googleapis.auth.oauth2.GoogleCredential(deprecated)` to `com.google.auth.oauth2.GoogleCredentials`

### DIFF
--- a/fiat-google-groups/src/main/java/com/netflix/spinnaker/fiat/roles/google/GoogleDirectoryUserRolesProvider.java
+++ b/fiat-google-groups/src/main/java/com/netflix/spinnaker/fiat/roles/google/GoogleDirectoryUserRolesProvider.java
@@ -16,7 +16,6 @@
 
 package com.netflix.spinnaker.fiat.roles.google;
 
-import com.google.api.client.googleapis.auth.oauth2.GoogleCredential;
 import com.google.api.client.googleapis.batch.BatchRequest;
 import com.google.api.client.googleapis.batch.json.JsonBatchCallback;
 import com.google.api.client.googleapis.json.GoogleJsonError;
@@ -32,16 +31,22 @@ import com.google.api.services.directory.Directory;
 import com.google.api.services.directory.DirectoryScopes;
 import com.google.api.services.directory.model.Group;
 import com.google.api.services.directory.model.Groups;
+import com.google.auth.http.HttpCredentialsAdapter;
+import com.google.auth.oauth2.GoogleCredentials;
+import com.google.auth.oauth2.ServiceAccountCredentials;
 import com.netflix.spinnaker.fiat.model.resources.Role;
 import com.netflix.spinnaker.fiat.permissions.ExternalUser;
 import com.netflix.spinnaker.fiat.roles.UserRolesProvider;
 import java.io.FileInputStream;
 import java.io.IOException;
+import java.util.ArrayDeque;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.Deque;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
@@ -53,8 +58,6 @@ import lombok.EqualsAndHashCode;
 import lombok.Setter;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.commons.lang3.StringUtils;
-import org.springframework.beans.PropertyAccessor;
-import org.springframework.beans.PropertyAccessorFactory;
 import org.springframework.beans.factory.InitializingBean;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
@@ -180,7 +183,7 @@ public class GoogleDirectoryUserRolesProvider implements UserRolesProvider, Init
     }
 
     try {
-      Groups groups = getGroupsFromEmail(userEmail);
+      Groups groups = getGroupsFromEmailRecursively(userEmail);
       if (groups == null || groups.getGroups() == null || groups.getGroups().isEmpty()) {
         return new ArrayList<>();
       }
@@ -189,6 +192,54 @@ public class GoogleDirectoryUserRolesProvider implements UserRolesProvider, Init
     } catch (IOException ioe) {
       throw new RuntimeException(ioe);
     }
+  }
+
+  /**
+   * Retrieves all Google Groups associated with a given email address, including both direct and
+   * indirect group memberships, if configured to do so.
+   *
+   * <p>This method first fetches the groups the user is directly a member of via {@link
+   * #getGroupsFromEmail(String)}. If the configuration allows expanding indirect groups (i.e.,
+   * nested groups), it recursively traverses each group's membership to collect nested groups.
+   *
+   * <p>The method avoids cycles and duplicate group processing by maintaining a set of already
+   * collected group emails.
+   *
+   * @param email The email address whose group memberships should be retrieved.
+   * @return A {@link Groups} object containing all the direct and (optionally) indirect group
+   *     memberships.
+   * @throws IOException If an error occurs while retrieving group information.
+   */
+  protected Groups getGroupsFromEmailRecursively(String email) throws IOException {
+    final Groups groups = getGroupsFromEmail(email);
+    if (groups == null
+        || groups.getGroups() == null
+        || groups.getGroups().isEmpty()
+        || !config.isExpandIndirectGroups()) {
+      return groups;
+    }
+    final Set<String> collectedGroup = new HashSet<>();
+    final Deque<String> stack = new ArrayDeque<>();
+    for (Group g : groups.getGroups()) {
+      stack.push(g.getEmail());
+      collectedGroup.add(g.getEmail());
+    }
+    while (!stack.isEmpty()) {
+      String nextEmail = stack.pop();
+      Groups subGroups = getGroupsFromEmail(nextEmail);
+      if (subGroups == null || subGroups.getGroups() == null || subGroups.getGroups().isEmpty()) {
+        continue;
+      }
+      for (Group g : subGroups.getGroups()) {
+        if (collectedGroup.contains(g.getEmail())) {
+          continue;
+        }
+        stack.push(g.getEmail());
+        groups.getGroups().add(g);
+        collectedGroup.add(g.getEmail());
+      }
+    }
+    return groups;
   }
 
   protected Groups getGroupsFromEmail(String email) throws IOException {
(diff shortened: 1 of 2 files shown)


fiat-google-groups/src/test/groovy/com/netflix/spinnaker/fiat/roles/google/GoogleDirectoryUserRolesProviderSpec.groovy
```diff
@@ -80,7 +80,77 @@ class GoogleDirectoryUserRolesProviderSpec extends Specification {
...
+            "child2@example.com",
+            ])
+        result.groups.size() == 2
+    }
+
     private static ExternalUser externalUser(String id) {
         return new ExternalUser().setId(id)
     }
-}
+}
```
dbyron-sf: missing end-of-line

fiat-google-groups/src/test/groovy/com/netflix/spinnaker/fiat/roles/google/GoogleDirectoryUserRolesProviderSpec.groovy
```diff
@@ -80,7 +80,77 @@ class GoogleDirectoryUserRolesProviderSpec extends Specification {
...
+        result.groups*.email.containsAll([
+            "child1@example.com",
+            "child2@example.com",
+            "grandchild1@example.com",
+            "grandchild2@example.com"
+            ])
+        result.groups.size() == 4
+    }
+
+        def "should not recursively collect all nested groups"() {
```
dbyron-sf: There's extra whitespace, and a lot of duplication here.  Can you try parameterizing the previous test?  Something like annotating with `@Unroll`, adding
```
where:
expandIndirectGroups << [false, true]
```
and adding `#expandIndirectGroups` to the test name and logic.

dbyron-sf:
Thanks for the PR @himanhsugusain!  Can you add some info about the impact of this change?  Like, before this PR a user who is a member of ABC groups isn't authorized for XYZ, but is afterwards (or whatever the impact actually is)?

dbyron-sf:
Also, since this is a feature, it'd be great to have the PR against [the monorepo](https://github.com/spinnaker/spinnaker).

himanhsugusain:
@dbyron-sf  Made the recommended fix to test. You can review once more.
Added more info on the change.
I did start with monorepo, but could not get my IDE setup to work with it properly ( too many things to build ).
Can you please cherry-pick this to monorepo after merge.
I'll try to raise next PRs to monorepo.

fiat-google-groups/src/main/java/com/netflix/spinnaker/fiat/roles/google/GoogleDirectoryUserRolesProvider.java
```diff
@@ -272,6 +311,13 @@ public static class Config {
     /** Google Apps for Work domain, e.g. netflix.com */
     private String domain;
 
+    /** expand indirect groups for emails */
+    private boolean expandIndirectGroups = false;
+
+    public boolean getExpandIndirectGroups() {
```
dbyron-sf: No need for this with `@Data`.  I believe the method lombok gives is called `isExpandIndirectGroups`.

fiat-google-groups/src/test/groovy/com/netflix/spinnaker/fiat/roles/google/GoogleDirectoryUserRolesProviderSpec.groovy
```diff
@@ -4,6 +4,21 @@ import com.netflix.spinnaker.fiat.permissions.ExternalUser
 import com.google.api.services.directory.model.Group;
 import com.google.api.services.directory.model.Groups;
 import spock.lang.Specification
+import spock.lang.Unroll
+
+
+class TestGoogleDirectoryUserRolesProvider extends GoogleDirectoryUserRolesProvider {
```
dbyron-sf: Can we do without this class by setting the appropriate config property in GoogleDirectoryUserRolesProvider?

fiat-google-groups/src/test/groovy/com/netflix/spinnaker/fiat/roles/google/GoogleDirectoryUserRolesProviderSpec.groovy
```diff
@@ -75,9 +90,48 @@ class GoogleDirectoryUserRolesProviderSpec extends Specification {
...
+        when:
+        def result = provider.getGroupsFromEmailRecursively("root@example.com")
 
-
+        then:
+        result.groups*.email.containsAll(groupsContent)
+        result.groups.size() == totalEmails
+
+        where:
+        expandIndirectGroups << [true, false]
```
dbyron-sf: Maybe the multi-column syntax is easier to read? Something like
```
expandIndirectGroups | totalEmails | groupsContent
----------------------------------------------------
true                 | 4           | ["child1@example.com", "child2@example.com", "grandchild1@example.com", "grandchild2@example.com"]
false.               | 2           | ["child1@example.com", "child2@example.com"]
```

fiat-google-groups/src/main/java/com/netflix/spinnaker/fiat/roles/google/GoogleDirectoryUserRolesProvider.java
himanhsugusain: ah! I checked for @Configuration not @Data.
Thanks

fiat-google-groups/src/main/java/com/netflix/spinnaker/fiat/roles/google/GoogleDirectoryUserRolesProvider.java
```diff
@@ -211,12 +246,15 @@ protected Groups getGroupsFromEmail(String email) throws IOException {
...
-      if (StringUtils.isNotEmpty(config.getCredentialPath())) {
-        return GoogleCredential.fromStream(new FileInputStream(config.getCredentialPath()));
+      if (StringUtils.isNotEmpty(config.getCredentialPath())
+          && StringUtils.isNotEmpty(config.getAdminUsername())) {
+        return ServiceAccountCredentials.fromStream(new FileInputStream(config.getCredentialPath()))
+            .createScoped(SERVICE_ACCOUNT_SCOPES) // add other scopes as needed
+            .createDelegated(config.getAdminUsername());
       } else {
-        return GoogleCredential.getApplicationDefault();
+        return ServiceAccountCredentials.getApplicationDefault();
```
jasonmcintosh: Hrmm, I don't know GCP stuff as well, but doesn't this more narrowly scope the credentials to a service account vs. a possibly more generic account credential?  I'm not sure that's good/bad RIGHT NOW... or even really whether a non service account scoped credential would have had THAT much of an impact.  Just noticing this vs. the GoogleCredential change...

fiat-google-groups/src/main/java/com/netflix/spinnaker/fiat/roles/google/GoogleDirectoryUserRolesProvider.java
```diff
@@ -253,6 +288,10 @@ private static Role toRole(Group g, Config.RoleSource src) {
     }
   }
 
+  protected boolean expandIndirectGroups() {
```
jasonmcintosh: NIT:  Since there's ALREADY access to the config option... having a protected method like this... seems to be redundent/not needed.  I'd inline this.  Reduces scope on these kinds of configs to ONLY be on the config object.

himanhsugusain: agreed.
I created this because I was having difficulty in accessing it from the test (skill issue xD ). Now that is fixed so this can go as well.

fiat-google-groups/src/main/java/com/netflix/spinnaker/fiat/roles/google/GoogleDirectoryUserRolesProvider.java
himanhsugusain: `ServiceAccountCredentials` just extends `GoogleCredentials` , so they are essentially the same. Still changing it to use `GoogleCredentials` for clarity.

fiat-google-groups/src/main/java/com/netflix/spinnaker/fiat/roles/google/GoogleDirectoryUserRolesProvider.java
```diff
@@ -191,6 +194,38 @@ public List<Role> loadRoles(ExternalUser user) {
     }
   }
 
+  protected Groups getGroupsFromEmailRecursively(String email) throws IOException {
```
dbyron-sf, line 213: javadoc please

himanhsugusain: Long live gpt.

fiat-google-groups/src/main/java/com/netflix/spinnaker/fiat/roles/google/GoogleDirectoryUserRolesProvider.java
```diff
@@ -211,12 +262,15 @@ protected Groups getGroupsFromEmail(String email) throws IOException {
...
   }
 
-  private GoogleCredential getGoogleCredential() {
+  private GoogleCredentials getGoogleCredential() {
     try {
-      if (StringUtils.isNotEmpty(config.getCredentialPath())) {
-        return GoogleCredential.fromStream(new FileInputStream(config.getCredentialPath()));
+      if (StringUtils.isNotEmpty(config.getCredentialPath())
+          && StringUtils.isNotEmpty(config.getAdminUsername())) {
+        return ServiceAccountCredentials.fromStream(new FileInputStream(config.getCredentialPath()))
```
jasonmcintosh, line 269: FYI:  Watch this as well... GoogleCredential vs. ServiceAccountCredential...

jasonmcintosh: That said... given the scoping... this MAY be fine.  This is one I don't know the GCP stuff AS WELL ... so asking :)

dbyron-sf (approved):

jasonmcintosh (approved):

